PR #4, opened 2017-09-13
Improve database caching between builds

### LINKED ISSUE #3: Improve database caching between builds
## Problem

Currently the database dump is downloaded and the SQL file is stored in the `.data` directory which is cached between builds. This is good, however when this file is large (i.e. several gigabytes) the process of importing the db can take a long time.

## Goal

To improve the speed of builds when a large database is in use.

## Proposed solution

The solution will need to support CircleCI 1.0.

* Enable `docker-engine` service on the build image.
* Replace the default local mysql service with one running inside a container.
* After the DB dump has been fetched, create a hash of the file to establish a version.
* If the hash is new, build a new mysql container image with the SQL dump pre-loaded. 
* Tag the new image with the version hash.
* Start the mysql container image.

... build proceeds as usual.

nicksantamaria:
Fixes #3 

## What does this PR do?

* Disables the default CircleCI mysql service and replaces with a docker container running mysql.
* Adds a command to database fetch step which executes a remote bash script. This bash script builds a container image with the database pre-loaded.
* Adds some environment variables to support this script.
* Removed 2 build steps from `composer.json` which dropped and re-imported the database dump. (note: this may need further input to accommodate local dev).

## Where should the reviewer start?

* Review the db container image build script - https://gist.github.com/nicksantamaria/43f44c45ac2cb31298cacd5c546902ff
* Review new environment variables in `circle.yml`
* Ensure workflow meets requirements.

## Background

The high-level approach for this feature was described in #3, however there were several implementation details which lead to this end result.

* The docker project are vehemently against Dockerfiles which rely on host-specific contexts. This killed the idea to have a `drupal-dev-containers` repo hosting the Dockerfiles, as the database dump needs to exist in the same directory at runtime. This was problematic as the process needed to be flexible enough for the db dump to be sourced in any number of ways (rsync, ftp, http etc..) and we do not want to have to support and maintain all those methods.
* The next idea to have a `docker run` >> `docker exec` >> `docker export` did not work as expected, as the exported "image" did not have all of the configuration provided by the base image `Dockerfile` (`ENV`, `ENTRYPOINT` etc..)
* The end solution was an evolution of the original idea, but with a dynamically generated Dockerfile. This means that the end-user can fetch the db dump however they wish, as long as it is stored in the right location prior to the db container being built.

## Anything else?

* As noted above, I removed a few lines from composer.json which will affect local dev. Looking for advice on best approach here.
* Once happy with the container build bash script, it should be moved to the `integratedexperts` namespace, and the location updated in this branch.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -33,17 +33,16 @@ dependencies:
     # Cache DB dumps directory.
     - .data
   pre:
-    # Stopping default mysql service so the db container can expose itself on port 3306.
-    - sudo service mysql stop
     # Download DB dump from public URL to test Drupal-dev project.
     # Remove the line below in your project.
     - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi
     # [META] Uncomment the line below to download DB dump from FTP, using
     # variables set in Circle CI UI.
     # - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -u $FTP_USER:$FTP_PASS "ftp://$FTP_HOST/db_d7.dist.sql" -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi
-    # Load db container image from cache directory.
+    - echo "Load db container image from cache directory"
     - docker load --input "${DB_CONTAINER_TAR}"
-    - docker run -d -p 3306:3306 "mysql/${DB_NAME}:latest"
+    - echo "replacing default mysql service with db container"
+    - sudo service mysql stop && docker run -d -p 3306:3306 "mysql/${DB_NAME}:latest"
     # Set a time zone to make sure that PHP's time functions work correctly.
     - echo "date.timezone = Australia/Sydney" >> /opt/circleci/php/$(phpenv global)/etc/php.ini
     # Disable xdebug to speed-up composer.
--- a/circle.yml
+++ b/circle.yml
@@ -30,6 +30,8 @@ dependencies:
     # Cache DB dumps directory.
     - .data
   pre:
+    # Stopping default mysql service so the db container can expose itself on port 3306.
+    - sudo service mysql stop
     # Download DB dump from public URL to test Drupal-dev project.
     # Remove the line below in your project.
     - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $BUILD_DIR/.data/db.dist.sql; fi
--- a/circle.yml
+++ b/circle.yml
@@ -21,6 +21,8 @@ machine:
     PATH: "$HOME/.config/composer/vendor/bin:$PATH"
   hosts:
     local.mysiteurl: 127.0.0.1
+  services:
+    - docker
 
 dependencies:
   cache_directories:
--- a/circle.yml
+++ b/circle.yml
@@ -36,10 +36,10 @@ dependencies:
     - sudo service mysql stop
     # Download DB dump from public URL to test Drupal-dev project.
     # Remove the line below in your project.
-    - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $BUILD_DIR/.data/db.dist.sql; fi
+    - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $DB_SOURCE_DUMP; bash <(curl -s https://gist.github.com/nicksantamaria/43f44c45ac2cb31298cacd5c546902ff); fi
     # [META] Uncomment the line below to download DB dump from FTP, using
     # variables set in Circle CI UI.
-    # - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -u $FTP_USER:$FTP_PASS "ftp://$FTP_HOST/db_d7.dist.sql" -o $BUILD_DIR/.data/db.dist.sql; fi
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -4,9 +4,12 @@ machine:
   environment:
     BUILD_DIR: $HOME/$CIRCLE_PROJECT_REPONAME
     DB_NAME: "circle_test"
-    DB_USER: "ubuntu"
+    DB_USER: "root"
```
AlexSkrypnyk, line 7: can this be retained as `ubuntu`?

circle.yml
```diff
@@ -4,9 +4,12 @@ machine:
...
     DB_PASS: ""
     DB_HOST: "127.0.0.1"
+    DB_SOURCE_DUMP: "$BUILD_DIR/.data/db.dist.sql"
+    DB_CONTAINER_TAR: "$BUILD_DIR/.data/db-container.tar"
+    DB_CONTAINER_SCRIPT: "https://gist.githubusercontent.com/nicksantamaria/43f44c45ac2cb31298cacd5c546902ff/raw/a1092fe8602546944aca961e83b303f95c44a9e6/db-container-build.sh"
```
AlexSkrypnyk: can you please replace this variable with in-place value of `https://goo.gl/eoeTMb` - this redirects to an actual code, but I can track it and replace if required

circle.yml
```diff
@@ -21,19 +24,26 @@ machine:
...
     - ~/.composer/cache
     # Cache DB dumps directory.
     - .data
   pre:
+    # Stopping default mysql service so the db container can expose itself on port 3306.
+    - sudo service mysql stop
     # Download DB dump from public URL to test Drupal-dev project.
     # Remove the line below in your project.
-    - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $BUILD_DIR/.data/db.dist.sql; fi
+    - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi
```
AlexSkrypnyk: can you please use multiline YAML syntax for this so that end user could easily enable/disable preferred way of DB handling in one place rather then scanning through many lines

```
- |
    if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi

...
```

circle.yml
```diff
@@ -21,19 +24,26 @@ machine:
...
     # [META] Uncomment the line below to download DB dump from FTP, using
     # variables set in Circle CI UI.
-    # - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -u $FTP_USER:$FTP_PASS "ftp://$FTP_HOST/db_d7.dist.sql" -o $BUILD_DIR/.data/db.dist.sql; fi
+    # - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -u $FTP_USER:$FTP_PASS "ftp://$FTP_HOST/db_d7.dist.sql" -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi
+    # Load db container image from cache directory.
+    - docker load --input "${DB_CONTAINER_TAR}"
```
AlexSkrypnyk: same as above - multiline YAML and group with previous so that there is a single point of modification

circle.yml
```diff
@@ -21,19 +24,26 @@ machine:
...
+    - sudo service mysql stop
     # Download DB dump from public URL to test Drupal-dev project.
     # Remove the line below in your project.
-    - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $BUILD_DIR/.data/db.dist.sql; fi
+    - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -L https://goo.gl/WFtJbT -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi
     # [META] Uncomment the line below to download DB dump from FTP, using
     # variables set in Circle CI UI.
-    # - if [ ! -f $BUILD_DIR/.data/db.dist.sql ] ; then mkdir $BUILD_DIR/.data && curl -u $FTP_USER:$FTP_PASS "ftp://$FTP_HOST/db_d7.dist.sql" -o $BUILD_DIR/.data/db.dist.sql; fi
+    # - if [ ! -f "${DB_CONTAINER_TAR}" ] ; then mkdir $BUILD_DIR/.data && curl -u $FTP_USER:$FTP_PASS "ftp://$FTP_HOST/db_d7.dist.sql" -o $DB_SOURCE_DUMP; bash <(curl -s "${DB_CONTAINER_SCRIPT}"); fi
+    # Load db container image from cache directory.
```
AlexSkrypnyk: pls covert thit to `echo "blah"` to make it verbose and self-documentting

composer.json
```diff
@@ -107,8 +107,6 @@
       "@build"
     ],
     "build-db": [
-      "drush @local.mysiteurl sql-drop -y",
```
AlexSkrypnyk, line 110: please revert this and add `[ "$CI" == "0" ] && drush @local.mysiteurl sql-drop -y`, `CI` env variable is populated in CI

circle.yml
```diff
@@ -4,9 +4,12 @@ machine:
   environment:
     BUILD_DIR: $HOME/$CIRCLE_PROJECT_REPONAME
     DB_NAME: "circle_test"
-    DB_USER: "ubuntu"
+    DB_USER: "root"
     DB_PASS: ""
     DB_HOST: "127.0.0.1"
+    DB_SOURCE_DUMP: "$BUILD_DIR/.data/db.dist.sql"
+    DB_CONTAINER_TAR: "$BUILD_DIR/.data/db-container.tar"
```
AlexSkrypnyk: please make this variable inline as the end user does not need to change it to make things work

nicksantamaria: I think it is useful to retain this environment variable as it is a parameter to the container build script. 

Even though the end-user doesn't have to change it, it means we have consistent paths between `circle.yml` and the bash script.

circle.yml
nicksantamaria: Yep, I've updated the script URL to `https://goo.gl/eoeTMb`

nicksantamaria:
I've refactored this a bit. There is now a feature toggle - changing the `DB_CONTAINER_ENABLE` environment variable will disable the entire feature (including in `composer.json`).

PR #586, opened 2020-05-20
Added support for core-recommended.

AlexSkrypnyk:
- Updated drush to `9 || 10`.
- Updated phpdotenv to `4`.
- `.gitattributes` - using scaffold version
- `.gitignore` - using scaffold version + DrevOps custom exclusions.
- Fixed configuration import when the site is built.
- Added unblocking of the admin user.


composer.json
```diff
@@ -75,50 +77,49 @@
             "Utilities\\composer\\DrupalSettings::create"
         ],
         "post-update-cmd": [
-            "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
```
AlexSkrypnyk, line 78: this is no longer required as `drupal/core-composer-scaffold` is now reacts on `composer update`

.editorconfig
```diff
@@ -1,14 +1,17 @@
-[*.js]
-indent_size = 2
-
-[composer.{json,lock}]
-indent_size = 4
+# Drupal editor configuration normalization
```
AlexSkrypnyk, line 1: from scaffold repo.

.gitattributes
```diff
@@ -1,3 +1,65 @@
+# Drupal git normalization
```
AlexSkrypnyk, line 1: from scaffold repo.

Issue #601, opened 2020-10-19
Add composer check-platform-reqs


AlexSkrypnyk:
add validation `composer check-platform-reqs`

AlexSkrypnyk:
this PR does not make sense - the validation happens when composer install runs at image build stage.

PR #605, opened 2020-10-30
[#601] Added checking of platform requirements on build.

AlexSkrypnyk:


AlexSkrypnyk:
this PR does not make sense - the validation happens when `composer install` runs at image build stage.

PR #627, opened 2021-06-09
Fixed environment detection in Lagoon.

AlexSkrypnyk:


richardgaunt (approved):
This was a great find, Alex

docroot/sites/default/settings.php
```diff
@@ -160,7 +160,7 @@
   }
   // Use a dedicated branch to identify production environment.
   // This is useful when 'production' Lagoon environment is not provisioned yet.
-  elseif (getenv('LAGOON_GIT_BRANCH') == getenv('LAGOON_PRODUCTION_BRANCH')) {
+  elseif (getenv('LAGOON_GIT_BRANCH') == getenv('LAGOON_PRODUCTION_BRANCH', 'production')) {
```
richardgaunt:
```suggestion
  elseif (getenv('LAGOON_GIT_BRANCH') !== FALSE && getenv('LAGOON_GIT_BRANCH') == getenv('LAGOON_PRODUCTION_BRANCH')) {
```

The second parameter does not set a default value it is a local_only flag: - Set to true to only return local environment variables (set by the operating system or putenv).

The suggested change ensures that a LAGOON_GIT_BRANCH has a value set before proceeding to check equality.

richardgaunt (changes requested):
The conditional we worked on doesn't work in the desired way, I've got a suggested change to have it work how we want.

richardgaunt (approved):
Thanks Alex, I've committed into VU already and it's detecting that it is not production correctly when there are no environmental variables.

PR #632, opened 2021-07-17
Remove port check from builds on linux

richardgaunt:
### Issue

Checking for whether Port 80 is occupied by a service other than Docker in `doctor.sh` fails incorrectly on Ubuntu (using 20.04). 

I usually turn off the check with `DOCTOR_CHECK_PORT=0` in .env but it would be great if I didn't need to do this each time.

This PR adds an OS check before proceeding with the port check.


AlexSkrypnyk (approved):

AlexSkrypnyk:
@richardgaunt Thank you for your contribution!

PR #651, opened 2021-08-13
[9.x] Porting D8 do D9.

AlexSkrypnyk:
- Ported D8 to D9
- Updated tests and test fixtures


.circleci/config.yml
```diff
@@ -203,11 +203,11 @@ jobs:
             #: Lookup cache based on the default branch and a timestamp. Allows
             #: to use cache from the very first build on the day (sanitized
             #: database dump, for example).
-            - v1-db8-{{ checksum "/tmp/DB_CACHE_BRANCH" }}-{{ checksum "/tmp/DB_CACHE_FALLBACK" }}-{{ checksum "/tmp/DB_CACHE_TIMESTAMP" }}
+            - v5-db9-{{ checksum "/tmp/DB_CACHE_BRANCH" }}-{{ checksum "/tmp/DB_CACHE_FALLBACK" }}-{{ checksum "/tmp/DB_CACHE_TIMESTAMP" }}
```
AlexSkrypnyk, line 206: here and below - had to increment the cache key to invalidate old DB docker images (if not invalidated - old images will be used in tests)

Once merged to `9.x` - the `v5` will be reverted to `v1`

.env
```diff
@@ -254,7 +254,7 @@ DREVOPS_COMMIT=
 #;< !DATABASE_DOWNLOAD_SOURCE_DOCKER_REGISTRY
 # Location of the demo database used for demonstration with CURL database
 # dump as a type of file source.
-CURL_DB_URL=https://raw.githubusercontent.com/wiki/drevops/drevops/db_d8.dist.sql.md
+CURL_DB_URL=https://raw.githubusercontent.com/wiki/drevops/drevops/db_d9.dist.sql.md
```
AlexSkrypnyk, line 257: this and below - would need to base the file name on `$DRUPAL_VERSION` so that these changes are no longer required

docroot/sites/default/default.services.yml
```diff
@@ -36,6 +36,22 @@ parameters:
     # @default none
     # cookie_domain: '.example.com'
     #
+    # Set the session ID string length. The length can be between 22 to 256. The
```
AlexSkrypnyk, line 39: here and below - these are updates from scaffolding files. These files are preserved here for refrerence

richardgaunt (approved):
This is fantastic work Alex. I've built and works without issue.

Just going to note that the demo DB does not activate:

1. Media and associated modules
2. Environment indicator
3. Redirect
4. Token
But this is not directly related to the PR

.env
richardgaunt: Is this PR the place to do it? Or are you noting for future PR / improvement.

AlexSkrypnyk:
> This is fantastic work Alex. I've built and works without issue.
> 
> Just going to note that the demo DB does not activate:
> 
> 1. Media and associated modules
> 2. Environment indicator
> 3. Redirect
> 4. Token
>    But this is not directly related to the PR

@richardgaunt 

Thank you for reviewing

The DB indeed does not enable those modules right now (for any versions of DrevOps). This is something that needs to be looked at later.

PR #657, opened 2021-09-02
Fixed drush ahoy command.

AlexSkrypnyk:


richardgaunt (approved):
Do you know the difference? Why does the old command work but not correctly work?

AlexSkrypnyk:
@richardgaunt 
not sure. Something could have change either in Docker (Compose) or in the CLI container itself. I do not have the time to investigate the actual cause, unfortunately.

richardgaunt:
No worries, I couldn't find the root cause either

PR #1087, opened 2023-11-08
Added Rector.

AlexSkrypnyk:


AlexSkrypnyk:
Need to add exclusions for `sites/simpletest` and `sites/default/files`

Issue #1213, opened 2024-02-29
Update Bats tests to use steps runner


AlexSkrypnyk:
Currently, bats tests for scripts are using assertions that are hard to maintain as they are using numbers.

Bats-helpers provides a helper to define multiple steps as an array, removing the need to maintain numerous assertions.

A [test for `provision.sh`](https://github.com/drevops/drevops/blob/develop/.drevops/tests/bats/provision.bats#L49) already has this implementation working.

Candidates for the update:
- https://github.com/drevops/drevops/blob/develop/.drevops/tests/bats/notify.bats

PR #1221, opened 2024-03-08
[ #1213] Updated tests for notify scripts to use the step runner format.

richardgaunt:
Partial implementation of #1213

## Checklist before requesting a review

- [x] I have formatted the subject to include ticket number as `[#123] Verb in past tense with dot at the end.`
- [x] I have added a link to the issue tracker
- [x] I have provided information in `Changed` section about WHY something was done if this was not a normal implementation
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have run new and existing relevant tests locally with my changes, and they passed
- [x] I have provided screenshots, where applicable

## Changed

1. Updated most tests to step_runner - note multi-line strings don't work with this helper.


.drevops/tests/bats/notify.bats
```diff
@@ -8,48 +8,75 @@ load _helper.bash
 
 @test "Notify: skip" {
   pushd "${LOCAL_REPO_DIR}" >/dev/null || exit 1
+  declare -a STEPS=(
```
AlexSkrypnyk: @richardgaunt 
Could you please revert this. I should have been more clear: we need to only convert tests that are using mocks. This one does not use mocks so it is easier to maintain it as normal asserts. sorry

AlexSkrypnyk (changes requested):
@richardgaunt 
Could you please revert the changes to tests that are not using mocks. We need to use `STEPS` only with tests that user the mocking system.

Sorry, I should have been more clear about the requirements.

PR #1235, opened 2024-03-16
Added unit tests for ssh key setup.

richardgaunt:
## Checklist before requesting a review

- [x] I have formatted the subject to include ticket number as `[#123] Verb in past tense with dot at the end.`
- [x] I have added a link to the issue tracker
- [ ] I have provided information in `Changed` section about WHY something was done if this was not a normal implementation
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run new and existing relevant tests locally with my changes, and they passed
- [ ] I have provided screenshots, where applicable

## Changed

1. Added tests for ssh setup.

### DIFF
--- a/.scaffold/tests/bats/deployment--artifact.bats
+++ b/.scaffold/tests/bats/deployment--artifact.bats
@@ -18,12 +18,6 @@ load _helper.deployment.bash
 
   assert_output_contains "Missing required value for DREVOPS_DEPLOY_TYPES. Must be a combination of comma-separated values (to support multiple deployments): code, docker, webhook, lagoon."
 
-  export DREVOPS_DEPLOY_ALLOW_SKIP=0
-  export DREVOPS_DEPLOY_TYPES="invalid"
-  run ahoy deploy
-  assert_failure
-  assert_output_contains "No deployments found for: ${DREVOPS_DEPLOY_TYPES}."
-
   popd >/dev/null
 }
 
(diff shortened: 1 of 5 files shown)


.scaffold/tests/bats/deployment--artifact.bats
```diff
@@ -0,0 +1,124 @@
...
+    "Finished SSH setup."
+  )
+  mocks="$(run_steps "setup")"
+  run scripts/drevops/setup-ssh.sh
+  assert_success
+  run_steps "assert" "${mocks[@]}"
+
+
+  # Finding key with matching fingerprint.
+  export DREVOPS_SSH_PREFIX="test"
```
richardgaunt: @AlexSkrypnyk I think this is a bug by the way
https://github.com/drevops/scaffold/pull/1235/files#diff-0dcfadf730b435edce46a713666041f22379292a90babf2cb949a4f7aa7bfaf7R115

Because we search for a matching ssh key only to then select a key that has the md5 signature in it's file name - doesn't make sense to me

richardgaunt: @AlexSkrypnyk Could you please review the style of this unit test - does this meet your requirements? And I will continue on the rest of the deployment unit tests if this meets what you had in mind

AlexSkrypnyk: @richardgaunt 
Thank you for working on this

We want to test every script in `scripts/drevops` dir
We then want to test all paths in each script, meaning that it would require multiple tests for the same script with different input parameters.

In this instance, where we are testing correct inclusion of the available SSH key, we would want to:
1. Cleanup current env variable values (use `unset VAR`)
2. Populate the env with test data: create valid SSH key (for positive test) or broken SSH key (for negative test) AND add the fingerprint in different formats.
3. Setup tests, which will mock commands used in the script itself (for example, `ssh-agent` does not need to be real and can be just a mock that returns a success code or  error code)
3. Call the script
4. Assert the output of the script by calling the test steps helper.


[UPDATE]

Just re-read the code of this PR again - the name of the file was a bit confusing. I think the SSH setup script should be using own test file.

The approach in the test itself looks really good. Please continue.

.scaffold/tests/bats/deployment--artifact.bats
```diff
@@ -0,0 +1,144 @@
...
+  export DREVOPS_DEPLOY_TYPES=""
+  run ahoy deploy
+  assert_failure
+
+  assert_output_contains "Missing required value for DREVOPS_DEPLOY_TYPES. Must be a combination of comma-separated values (to support multiple deployments): code, docker, webhook, lagoon."
+
+  popd >/dev/null
+}
+
+@test "SSH Key Setup script" {
```
AlexSkrypnyk: This would be 
```
"Valid SSH Key, MD5 fingerprint" {
```

Then, there would be other cases:
- `"Valid SSH Key, SHA256 fingerprint"` {`
- `"Invalid SSH Key, MD5 fingerprint"` {`
- `"Invalid SSH Key, SHA256 fingerprint"` {`
- `"Valid SSH Key, SHA256 fingerprint, file from env var"` {` - in this case, any discovered keys with provided fingerprints would not be use and the key file name from the env variable will be used instead.

.scaffold/tests/bats/deployment--artifact.bats
```diff
@@ -0,0 +1,144 @@
...
+  mocks="$(run_steps "setup")"
+  run scripts/drevops/setup-ssh.sh
+  assert_success
+  run_steps "assert" "${mocks[@]}"
+
+  # Assert does not have key loaded
+  export DREVOPS_SSH_PREFIX="test"
+  export DREVOPS_test_SSH_FINGERPRINT="TEST"
+  export CI="1"
+  local file="${SSH_KEY_FIXTURE_DIR}/id_rsa_${DREVOPS_test_SSH_FINGERPRINT}"
```
AlexSkrypnyk: `DREVOPS_TEST_SSH_FINGERPRINT`

.scaffold/tests/bats/deployment--artifact.bats
richardgaunt: ok nice i will adapt it to this method

richardgaunt:
@AlexSkrypnyk This can be progressively merged as we go. Two questions I have though:
- Do we want a separate code coverage for these unit tests and so we can monitor the coverage of these tests?
- Do we want to group these tests in their own directory to keep them separate from the integration tests?

AlexSkrypnyk (approved):
@richardgaunt 
thank you for working on these. I'm going to merge them now.

We will move tests around a bit later, once we have more unit tests as this may require splitting some of the existing bats tests into unit and functional ones (and it is not that easy to do ).

PR #1334, opened 2024-06-19
Update Container images to v24.6.0.

AlexSkrypnyk:


composer.json
```diff
@@ -148,7 +148,12 @@
         "patchLevel": {
             "drupal/core": "-p2"
         },
-        "patches": {}
+        "patches": {
```
AlexSkrypnyk: @tannguyen04 
please find another way to not include these patches

tannguyen04: Hi @AlexSkrypnyk 
New release of search_api_solr is able to fix the issue. No need patches. 
https://www.drupal.org/project/search_api/issues/3454939#comment-15645298

PR #1369, opened 2024-08-27
Updated Lagoon deployment script to use long-form options.

AlexSkrypnyk:


.scaffold/tests/bats/deployment1.bats
```diff
@@ -63,7 +63,7 @@ load _helper.deployment.bash
 
   # Make sure that one of the excluded directories will be ignored in the
   # deployment artifact.
-  mkdir -p "${SRC_DIR}"/web/themes/custom/star_wars/node_modules
+  mkdir --project "${SRC_DIR}"/web/themes/custom/star_wars/node_modules
```
richardgaunt: This doesnt make sense

`mkdir -p` makes directory and all intermediate directories nothing to do with `--project`

richardgaunt (approved):
Thanks yes is much nicer to see whats going on

PR #1619, opened 2025-05-05
Minor installer fixes.

skipper-vp:
- Changed wording for utfPos() to isAsciiStart() because this method checks if the string starts with Ascii.
- Removed duplicate from .vortex/installer/src/Utils/File.php
- Fixed typo in OPTION_NO_ITERACTION
- Made $destination optional in protected function `downloadFromRemote(string $repo, string $ref, ?string $destination): string {` because it is optional in the calling method `public function download(string $repo, string $ref, ?string $dst = NULL): string {`

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected a typo in the "no-interaction" command-line option, ensuring it works as expected during installations.
	- Removed a duplicate entry from the internal file paths list to prevent redundancy.
- **Refactor**
	- Renamed a method for checking if a string starts with an ASCII character, improving code clarity.
	- Updated internal logic to use the new method name for ASCII detection in text processing.
	- Allowed destination paths to be nullable in certain download operations, enforcing non-null destinations.
- **Tests**
	- Updated test method names and logic to align with the renamed ASCII detection method.

### DIFF
--- a/.vortex/installer/src/Command/InstallCommand.php
+++ b/.vortex/installer/src/Command/InstallCommand.php
@@ -29,7 +29,7 @@ class InstallCommand extends Command {
 
   const OPTION_ROOT = 'root';
 
-  const OPTION_NO_ITERACTION = 'no-interaction';
+  const OPTION_NO_INTERACTION = 'no-interaction';
 
   const OPTION_CONFIG = 'config';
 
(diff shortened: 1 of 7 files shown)


skipper-vp:
@AlexSkrypnyk Please have a look and let me know if you'd like me to cover this with additional tests.

AlexSkrypnyk:
@coderabbitai full review

PR #2465, opened 2026-04-20
[#2441] Fixed Acquia DB download to skip auth header on S3 redirect.

### LINKED ISSUE #2441: Fix DB download from Acquia
- do not path authentication header to the final S3 download URL (the auth header is required only to query Acquia API)
- do not remove the downloaded file that cannot be opened - leave it for inspection.

AlexSkrypnyk:
Closes #2441

## Summary

Fixed two bugs in `scripts/vortex/download-db-acquia.sh` that caused Acquia DB downloads to fail.

The Acquia API `actions/download` endpoint returns a 302 redirect to a presigned S3 URL. The previous implementation followed the redirect with `-L`, which could send the `Authorization: Bearer` header to S3 - S3 presigned URLs reject requests that carry extra auth headers, causing the download to return an error body instead of the archive. The fix captures only the redirect URL (without following it) and then makes a separate, auth-free curl call to S3.

Additionally, error branches that deleted partially-downloaded or corrupt files with `rm -f` were changed to leave those files in place so they can be inspected when something goes wrong.

### `scripts/vortex/download-db-acquia.sh`

- **Auth header leak fix**: replaced `curl --progress-bar -L ... | extract_json_value "url"` with `curl -s -o /dev/null -w "%{redirect_url}" ...` to capture the S3 redirect URL without following it and without forwarding the `Authorization` header. The JSON-parsing branch was removed because the API only ever responds with a redirect, not a JSON body.
- **Preserve files on failure**: removed three `rm -f` calls from error branches (empty/missing download, invalid gzip, decompression failure) so the corrupt file is kept for inspection.
- **Move cleanup to success path**: moved `rm "${file_name_compressed}"` to after the decompression validity check so the compressed file is only deleted on a fully successful run.

### `.vortex/tests/bats/unit/download-db-acquia.bats`

- Updated three test-case curl mocks for the "Discovering backup URL" step to match the new `curl -s -o /dev/null -w %{redirect_url}` signature and to return a plain URL string instead of a JSON body.

## Before / After

**OLD flow**

```
Acquia API call
  curl --progress-bar -L
       -H "Authorization: Bearer <token>"
       https://…/actions/download
    │
    │  302 redirect (followed by -L)
    ▼
  S3 presigned URL  ← Authorization header forwarded here
    │
    │  S3 rejects extra auth → returns error XML body
    ▼
  Corrupt/empty file saved locally
    │
    │  gzip check or size check fails
    ▼
  rm -f  ← evidence deleted, nothing to inspect
```

**NEW flow**

```
Acquia API call (capture redirect only)
  curl -s -o /dev/null -w "%{redirect_url}"
       -H "Authorization: Bearer <token>"
       https://…/actions/download
    │
    │  302 redirect URL extracted, connection closed
    ▼
  backup_url = "https://s3.amazonaws.com/…?X-Amz-Signature=…"

S3 download (no auth header)
  curl --progress-bar -L "${backup_url}"
    │
    │  S3 accepts presigned URL, streams archive
    ▼
  Valid .sql.gz file saved locally
    │
    ├─ success → gunzip → rm compressed file
    │
    └─ failure → files left in place for inspection
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error diagnostics for Acquia database backup operations by preserving intermediate files for inspection when download or decompression fails, enabling improved troubleshooting
  * Optimized backup URL discovery mechanism

* **Tests**
  * Updated test cases for Acquia backup URL discovery functionality

### DIFF
--- a/.vortex/tests/bats/unit/download-db-acquia.bats
+++ b/.vortex/tests/bats/unit/download-db-acquia.bats
@@ -35,7 +35,7 @@ load ../_helper.bash
 
     # Mock backup URL curl call with its message
     "Discovering backup URL."
-    '@curl --progress-bar -L -H Accept: application/json, version=2 -H Authorization: Bearer test-token https://cloud.acquia.com/api/environments/env-id-456/databases/testdb/backups/backup-id-789/actions/download # {"url":"https://backup.example.com/db.sql.gz","expires":"2024-01-01T01:00:00+00:00"}'
+    '@curl -s -o /dev/null -w %{redirect_url} -H Accept: application/json, version=2 -H Authorization: Bearer test-token https://cloud.acquia.com/api/environments/env-id-456/databases/testdb/backups/backup-id-789/actions/download # https://backup.example.com/db.sql.gz'
 
     # Mock file download curl call with its message and side effect to create zipped archive
     "Downloading DB dump into file .data/testdb_backup_backup-id-789.sql.gz."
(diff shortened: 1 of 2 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.53% (201/204)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.53% (201/204)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.53% (201/204)
```
<details>

```
```
</details>

PR #2622, opened 2026-06-16
[#2621] Fixed CI 'Export DB' step not copying the processed dump back to the host.

### LINKED ISSUE #2621: CI 'Export DB' step writes the dump inside the container volume and never copies it back to the host
## Summary

In the CI database job, the `Export DB` step copies the downloaded `db.sql` into the `cli` container, imports it, then re-exports it with `./vendor/drevops/vortex-tooling/src/export-db db.sql`. That export runs inside the container and writes to `/app/.data/db.sql`. In CI `/app` is a Docker named volume (the host bind mount is stripped by the `Process the codebase to run in CI` step), so the re-exported dump stays inside the container volume. The step copies the dump **in** (host to container) but never **out** (container to host). As a result, the processed dump never reaches the runner's `.data`, the `Save DB cache` step cannot see it, and any modification made to the database just before the export is silently discarded.

## Affected files (repo `drevops/vortex`)

- `.github/workflows/build-test-deploy.yml` - `Export DB` step (around lines 308-316), `Save DB cache` step (around lines 322-327), `Process the codebase to run in CI` step (around line 259).
- `docker-compose.yml` - volume definitions (the `### removed in CI` / `## uncommented in CI` pattern).
- `vendor/drevops/vortex-tooling/src/export-db` and `export-db-file` (the export runs inside the container via `docker compose exec cli`).

Scope: file-based database provisioning (`VORTEX_PROVISION_TYPE=database`, no `VORTEX_EXPORT_DB_IMAGE`/`VORTEX_DB_IMAGE`). The container-image database path is unaffected.

## Root cause

The `Export DB` step (verbatim):

```yaml
- name: Export DB
  run: |
    if [ ! -f /tmp/download-db-success ]; then echo "==> Database download semaphore file is missing. DB export will not proceed."; exit 0; fi
    ./vendor/drevops/vortex-tooling/src/login-container-registry
    docker compose up --detach && sleep 15
    docker compose exec cli mkdir -p .data && docker compose cp -L .data/db.sql cli:/app/.data/db.sql
    docker compose exec cli bash -c "VORTEX_PROVISION_POST_OPERATIONS_SKIP=1 ./vendor/drevops/vortex-tooling/src/provision"
    ./vendor/drevops/vortex-tooling/src/export-db db.sql
  timeout-minutes: 30
```

1. In CI, `Process the codebase to run in CI` rewrites every `docker-compose.yml` with `sed -i -e '/###/d'` and `sed -i -e 's/##//'`. That deletes the `.:/app` host bind mount and enables the `app:/app` named volume. Inside the `cli` container, `/app` (and therefore `/app/.data`) is a container-private volume, not the runner's working directory.
2. Because there is no host bind mount, the step explicitly copies the downloaded dump **into** the container: `docker compose cp -L .data/db.sql cli:/app/.data/db.sql`.
3. `./vendor/drevops/vortex-tooling/src/export-db db.sql` is a host-side router. With no DB image configured it delegates to `export-db-file` via `docker compose exec -T cli ...`, so the actual `drush sql:dump` runs **inside** the container and writes `/app/.data/db.sql` into the named volume.
4. Nothing copies that file back out. The runner's `.data/db.sql` is still the byte-for-byte downloaded dump.
5. `Save DB cache` caches the runner's `.data` and is gated on `if: env.db_hash != hashFiles('.data')`, where `db_hash` was captured right after download. Since the runner's `.data` is unchanged by the `Export DB` step, the condition is not satisfied and the processed dump is not persisted.

The copy-in at step 2 shows the named-volume boundary is already understood; only the symmetric copy-out is missing.

## Impact

- Any database transformation performed between import and export is discarded in CI: sanitization beyond what `provision` does, cache-table truncation, anonymization, row pruning, dump-size reduction. It runs against the in-container database and the result never leaves the container.
- The dump produced by the `Export DB` step is never written to the host `.data`, so the `Save DB cache` condition `env.db_hash != hashFiles('.data')` is never satisfied by this step in CI. The intended "import, process, re-export, cache the processed dump" flow does not persist its output.
- Downstream jobs and the saved cache operate on the raw downloaded dump, not the processed one.

Locally this is hidden, because the `.:/app` bind mount is present, so the same in-container `export-db-file` write surfaces directly on the host. The divergence only appears in CI, where the bind mount is removed.

## How to observe

1. In the `Export DB` step, modify the database after `provision` and before `export-db` (for example `drush sql:query "TRUNCATE TABLE cache_dynamic_page_cache;"`).
2. Add `ls -Al .data` after the export. The listed size on the runner is unchanged from the downloaded dump.
3. Confirm with `docker compose exec -T cli ls -Al /app/.data` - the in-container dump reflects the modification, proving the export wrote into the volume rather than to the host.

## Proposed fix

Mirror the existing copy-in with a copy-out, immediately after the export, in the `Export DB` step:

```yaml
    ./vendor/drevops/vortex-tooling/src/export-db db.sql
    # In CI, /app is a named volume, so the in-container export must be
    # copied back to the runner to become the cached dump.
    docker compose cp -L cli:/app/.data/db.sql .data/db.sql
```

After this, the runner's `.data/db.sql` holds the processed dump, the `Save DB cache` condition is satisfied, and downstream jobs use the processed dump.

## Alternatives considered

- Handle the copy-out inside `export-db` (the host-side router). It already shells into the container for `export-db-file`, so it could `docker compose cp` the result back afterwards. This couples a generic tool to a CI-specific volume layout and would be redundant in bind-mounted contexts, so the workflow step is the better place; the copy-in already lives there.
- Keep a host bind mount for `.data` in CI. This reintroduces host-mount coupling that the named-volume CI mode deliberately removes, and is broader than the problem.

AlexSkrypnyk:
Closes #2621

## Summary

In CI the host bind mount (`.:/app`) is stripped and `/app` becomes a named Docker volume, so anything written inside the `cli` container after startup is invisible on the runner. The `Export DB` step copied the downloaded dump into the container, provisioned, and re-exported it to `/app/.data/db.sql` - but never copied that processed dump back out. The dump stayed in the named volume and was silently discarded, so `Save DB cache` cached the original raw download instead of the provisioned result, and any transformation applied between import and export was lost. The fix mirrors the existing copy-in with a copy-out after `export-db`, in both GitHub Actions and CircleCI.

## Changes

- `.github/workflows/build-test-deploy.yml` - added `docker compose cp -L cli:/app/.data/db.sql .data/db.sql` after `export-db` so the processed dump reaches the runner before the cache step. The GHA `database` job is file-based only (its copy-in has no fallback), so the copy-out is unconditional.
- `.circleci/config.yml` - added the same copy-out in the `database` job, guarded to run only when the dump file actually exists in the container: `docker compose exec -T cli test -f /app/.data/db.sql && docker compose cp ... || true`. The CircleCI job also supports database-in-image, where `export-db` builds an image and writes no `/app/.data/db.sql`; the existence check skips the copy-out in that case and is correct whether image mode is selected via an environment variable or `.env`.
- `.circleci/vortex-test-common.yml` - regenerated the derived test config so it stays in sync with `.circleci/config.yml` (checked by `lint.ci.sh`).
- `.vortex/installer/tests/Fixtures/handler_process/**` - regenerated installer snapshot fixtures: the `_baseline` and `*_circleci` scenarios gain the copy-out line; GHA scenario patch fixtures get mechanical hunk-offset shifts.

## Before / After

```
                         BEFORE

  runner host                    cli container (/app = named volume)
  ─────────────────              ─────────────────────────────────────
  .data/db.sql (raw)  ─copy IN─▶  /app/.data/db.sql (raw)
                                          │
                                    provision + export
                                          │
                                          ▼
                                  /app/.data/db.sql (processed)
                                         [stuck in named volume]

  .data/db.sql (raw) ◀─ NO copy-out ─── [lost]
       │
  Save DB cache ──▶ caches RAW dump  ✗


                         AFTER

  runner host                    cli container (/app = named volume)
  ─────────────────              ─────────────────────────────────────
  .data/db.sql (raw)  ─copy IN─▶  /app/.data/db.sql (raw)
                                          │
                                    provision + export
                                          │
                                          ▼
                                  /app/.data/db.sql (processed)
                                          │
  .data/db.sql       ◀─ copy OUT ─────────┘
  (processed)
       │
  Save DB cache ──▶ caches PROCESSED dump  ✓
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -285,6 +285,9 @@ jobs:
             docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c "VORTEX_PROVISION_POST_OPERATIONS_SKIP=1 ./vendor/drevops/vortex-tooling/src/provision"
             grep -q ^VORTEX_DB_IMAGE .env && rm .data/db.sql || true
             ./vendor/drevops/vortex-tooling/src/export-db db.sql
+            # In CI, /app is a named volume, so the in-container export must be
+            # copied back to the host to become the cached dump.
+            grep -q ^VORTEX_DB_IMAGE .env || docker compose cp -L cli:/app/.data/db.sql .data/db.sql
           no_output_timeout: 30m
 
       - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -285,7 +285,7 @@ jobs:
             docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c "VORTEX_PROVISION_POST_OPERATIONS_SKIP=1 ./vendor/drevops/vortex-tooling/src/provision"
             grep -q ^VORTEX_DB_IMAGE .env && rm .data/db.sql || true
             ./vendor/drevops/vortex-tooling/src/export-db db.sql
-            grep -q ^VORTEX_DB_IMAGE .env || docker compose cp -L cli:/app/.data/db.sql .data/db.sql
+            docker compose exec -T cli test -f /app/.data/db.sql && docker compose cp -L cli:/app/.data/db.sql .data/db.sql || true
           no_output_timeout: 30m
 
       - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -285,8 +285,6 @@ jobs:
             docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c "VORTEX_PROVISION_POST_OPERATIONS_SKIP=1 ./vendor/drevops/vortex-tooling/src/provision"
             grep -q ^VORTEX_DB_IMAGE .env && rm .data/db.sql || true
             ./vendor/drevops/vortex-tooling/src/export-db db.sql
-            # In CI, /app is a named volume, so the in-container export must be
-            # copied back to the host to become the cached dump.
             grep -q ^VORTEX_DB_IMAGE .env || docker compose cp -L cli:/app/.data/db.sql .data/db.sql
           no_output_timeout: 30m
 
--- a/.circleci/vortex-test-common.yml
+++ b/.circleci/vortex-test-common.yml
@@ -168,6 +168,9 @@ jobs:
             docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c "VORTEX_PROVISION_POST_OPERATIONS_SKIP=1 ./vendor/drevops/vortex-tooling/src/provision"
             grep -q ^VORTEX_DB_IMAGE .env && rm .data/db.sql || true
             ./vendor/drevops/vortex-tooling/src/export-db db.sql
+            # In CI, /app is a named volume, so the in-container export must be
+            # copied back to the host to become the cached dump.
+            grep -q ^VORTEX_DB_IMAGE .env || docker compose cp -L cli:/app/.data/db.sql .data/db.sql
           no_output_timeout: 30m
 
       - save_cache:
--- a/.circleci/vortex-test-common.yml
+++ b/.circleci/vortex-test-common.yml
@@ -168,7 +168,7 @@ jobs:
             docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c "VORTEX_PROVISION_POST_OPERATIONS_SKIP=1 ./vendor/drevops/vortex-tooling/src/provision"
             grep -q ^VORTEX_DB_IMAGE .env && rm .data/db.sql || true
             ./vendor/drevops/vortex-tooling/src/export-db db.sql
-            grep -q ^VORTEX_DB_IMAGE .env || docker compose cp -L cli:/app/.data/db.sql .data/db.sql
+            docker compose exec -T cli test -f /app/.data/db.sql && docker compose cp -L cli:/app/.data/db.sql .data/db.sql || true
           no_output_timeout: 30m
 
       - save_cache:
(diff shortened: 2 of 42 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (205/208)
```
<details>

```
```
</details>

PR #2692, opened 2026-06-22
[#2636] Deferred CI dev-dependency install until after the database cache is saved.

### LINKED ISSUE #2636: Defer CI dev-dependency install until after the database cache is captured
## Summary

Defer the `Install development dependencies` step in the CI `test` job until **after** the database cache is captured (saved), so the cache is stored before the slow composer-dev + yarn install rather than after it.

This is a follow-up optimization to #2627, which folded the database download + first-build-of-day caching into the `test` job. In #2627 the dev-dependency install currently runs before the cache is captured; reordering it is deferred to here.

## Why it is safe

The cache capture (`Download DB` + the `Export DB after download` step: a raw import via `provision` with `VORTEX_PROVISION_POST_OPERATIONS_SKIP=1`, then `export-db`, then copying the processed dump back to the host) needs only the **prod** dependencies - `drush` and `drevops/vortex-tooling` - which are already baked into the cli image by `.docker/cli.dockerfile` (`composer install --no-dev ...`) and present the moment `docker compose up` finishes. The dev dependencies (PHPUnit/Behat) and yarn packages are only needed by the test steps. The old standalone `database` job captured this exact cache with no `composer install` step at all, confirming dev deps are not a prerequisite.

## Benefits

- The cache lands sooner on the first build of the day (before the ~1-3 min dev-deps + yarn install), so concurrent/later builds that day pick it up faster.
- The cache survives downstream failures: if `composer install`/`yarn install` (or provision/tests) flake, the expensive download + export is already persisted instead of being thrown away and re-downloaded next run.

## Target order

```
Build stack -> Download DB -> Export DB (capture) -> Save cache -> Install dev deps -> Provision -> Tests
```

## Provider notes

- **GitHub Actions** (`.github/workflows/build-test-deploy.yml`): straightforward. The save is already gated by `if: matrix.instance == 0` (no file deletion), so just move `Install development dependencies` to after `Save DB cache`.
- **CircleCI** (`.circleci/config.yml`): needs a decision. The current node-0 save guard works by deleting `.data` on non-primary runners right before `save_cache`, and that runs *after* provision because provision still needs `.data` - both the main dump on a cache hit and the migration `db2.sql` on every cache-miss run. To move the save before provision, the guard has to change. Options:
  1. **Write-once cache** - both runners run `save_cache` with the same key; CircleCI keeps only the first write. Simplest, makes both providers identical, cache survives all later failures. Cost: both runners upload on the day's first build.
  2. **Explicit guard via stash** - stash `.data` aside on non-primary runners around `save_cache`, then restore it for provision. Preserves single-runner upload, keeps migration safe; adds two stash/restore steps.
  3. **Minimal** - leave the save after provision; only move `Install development dependencies` after it. Smallest change, but CircleCI then provisions on the baked prod image before dev deps are installed (asymmetric with GHA, and the cache is not saved if provision fails).

Regenerate `.circleci/vortex-test-common.yml` and the installer fixtures after the change.

AlexSkrypnyk:
Closes #2636

## Summary

Moved the `save_cache` step (and its guard logic) in the CI `test` job to run immediately after the database dump is exported, before `composer install`, `yarn install`, and `provision`. This ensures the cached DB dump lands sooner and is not lost if a later step fails. On GitHub Actions the change is a straight reorder since the `Save DB dump to cache` step is already gated to matrix instance 0 by a `if:` condition. On CircleCI, where `save_cache` has no per-node runtime condition, two parallel runners storing the same key concurrently would fail the build; the fix stashes `.data` aside on non-primary nodes immediately before `save_cache` (using a guarded `mv`) and restores it after, so only the primary runner writes to the cache while all runners retain the dump for `provision`.

## Changes

**`.github/workflows/build-test-deploy.yml`** - Moved the `Install development dependencies` step to after the existing `Save DB dump to cache` block; no guard logic needed as the `if: matrix.parallel-runner == 0` condition already restricts the save to a single runner.

**`.circleci/config.yml` and `.circleci/vortex-test-common.yml`** - Removed `Install development dependencies` from before `export-db`; inserted `Stash DB on non-primary runners so only the primary saves the cache` (conditional `mv .data .data.nosave`) and `save_cache` before `provision`; added `Restore stashed DB on non-primary runners for provisioning` (conditional `mv .data.nosave .data`) after `save_cache`; moved `Install development dependencies` to after the restore and before `provision`. Removed the now-superseded `Keep DB cache only on the primary parallel runner` step that destructively deleted `.data` post-provision.

**`.circleci/vortex-test-common.yml`** - Regenerated from the canonical `config.yml` to keep it in sync (mirrors the same reordering).

**`.vortex/installer/tests/Fixtures/`** - All installer snapshot fixtures were regenerated to reflect the new step order in both GitHub Actions and CircleCI variants.

## Before / After

```
GitHub Actions — test job step order
─────────────────────────────────────────────────────────────────────────
BEFORE                                  AFTER
──────────────────────────────────      ──────────────────────────────────
Build stack                             Build stack
│                                       │
▼                                       ▼
Install development dependencies  ──►  [moved to after cache save ─────►]
│                                       │
▼                                       ▼
Export / capture DB dump                Export / capture DB dump
│                                       │
▼                                       ▼
Save DB dump to cache                   Save DB dump to cache
(if: matrix.parallel-runner == 0)       (if: matrix.parallel-runner == 0)
│                                       │
▼                                       ▼
Provision site                          Install development dependencies
                                        │
                                        ▼
                                        Provision site

CircleCI — test job step order (parallel runners)
─────────────────────────────────────────────────────────────────────────
BEFORE                                  AFTER
──────────────────────────────────      ──────────────────────────────────
Build stack                             Build stack
│                                       │
▼                                       ▼
Install development dependencies  ──►  [moved to after cache save ─────►]
│                                       │
▼                                       ▼
Export / capture DB dump                Export / capture DB dump
│                                       │
▼                                       ▼
Keep DB cache only on primary           Stash .data on non-primary nodes
  (rm -rf .data on non-primary)           (mv .data .data.nosave, guarded)
│                                       │
▼                                       ▼
save_cache                              save_cache
│                                       │
▼                                       ▼
Provision site                          Restore stashed .data on
                                          non-primary nodes
                                          (mv .data.nosave .data, guarded)
                                        │
                                        ▼
                                        Install development dependencies
                                        │
                                        ▼
                                        Provision site

Key change on CircleCI: the old approach deleted .data so non-primary
runners had nothing to provision with; the new approach stashes and
restores it, so every runner provisions correctly.
```

## Summary by CodeRabbit

* **Chores**
  * Optimized CI/CD pipeline configuration for improved build reliability and efficiency during parallel testing operations.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -276,16 +276,6 @@ jobs:
           name: Build stack
           command: docker compose up --detach && docker builder prune --all --force
 
-      - run:
-          name: Install development dependencies
-          command: |
-            docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c " \
-              if [ -n \"${PACKAGE_TOKEN:-}\" ]; then export COMPOSER_AUTH='{\"github-oauth\": {\"github.com\": \"${PACKAGE_TOKEN-}\"}}'; fi && \
-              COMPOSER_MEMORY_LIMIT=-1 composer --ansi install --prefer-dist"
-            #;< TOOL_JEST
-            docker compose exec $(env | cut -f1 -d= | sed 's/^/-e /') -T cli bash -c "yarn install --frozen-lockfile"
-            #;> TOOL_JEST
-
       #;< !PROVISION_TYPE_PROFILE
       # Execute commands after the database download finished: if a fresh DB
       # dump was downloaded (cache miss), import it and export it back to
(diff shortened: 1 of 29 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.60% (211/214)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.60% (211/214)
```
<details>

```
```
</details>

PR #2867, opened 2026-07-30
[#2868] Pinned Hadolint to v2.15.0 and replaced its global ignores with justified per-line ones.

### LINKED ISSUE #2868: Pin the 'hadolint' image tag and replace its global rule suppressions
## Summary

Hadolint is invoked without an image tag in CI, so it resolves to `hadolint/hadolint:latest`. Any hadolint release can therefore turn a green pipeline red on a commit that changed nothing, on a default branch, without review. This is not hypothetical: it happened on 2026-07-30, and it broke the `lint` job in at least one consumer project built from this template.

The sibling linters in the same job are already pinned (`zavoloklom/dclint:3.1.0`, `ghcr.io/gitleaks/gitleaks:v8.30.1`), so hadolint is the outlier rather than the convention.

Resolving the fallout surfaced a second, larger problem in the same area: the shipped Dockerfiles suppress hadolint rules file-wide, and several of those suppressions no longer suppress anything. Both are in scope here, since a version pin without an audit of what that version is being told to ignore only freezes the blind spot in place.

### What happened

hadolint `v2.15.0` was published on 2026-07-30 at 13:39 UTC. It introduces two new rules that the template's shipped Dockerfiles trip:

- `DL3066` - Non-numeric user-id may not be resolvable by host system (severity `info`).
- `DL3064` - Potentially sensitive data should not be used in the `ARG` or `ENV` commands (severity `warning`).

hadolint's default failure threshold is `info`, so both severities fail the build. `v2.14.0` reported zero findings on the same files; `v2.15.0` reports 8. Nothing in the repository changed between the passing and failing runs - only the image that `latest` pointed at.

### Where the invocation is unpinned

Three places run hadolint without a tag:

- `.github/workflows/build-test-deploy.yml`
- `.circleci/config.yml`
- `.vortex/tests/lint.dockerfiles.sh`

All three should pin the same version so the GitHub Actions and CircleCI providers cannot drift apart, and so a local run of the test script matches what CI does.

### Global rule suppressions hide future violations

Five of the six shipped Dockerfiles carry a `# hadolint global ignore=` header. A global pragma disables the rule for the entire file, including for instructions added later, so a suppression added to excuse one known line silently blesses every future violation of that rule in that file.

This matters most for `DL3064`, which exists to catch secrets in `ARG` and `ENV`. Suppressed file-wide in `.docker/cli.dockerfile`, a future `ARG AWS_SECRET_ACCESS_KEY` in that file would pass the linter unnoticed. The same reasoning applies to `DL3008` and `DL3018`, which stop flagging unpinned packages in any instruction added after the pragma.

Suppressions should sit on the line they apply to, and each should state why. hadolint supports `# hadolint ignore=RULE # reason`, so a justified per-line suppression costs one line and no extra vertical space.

### Some suppressions no longer suppress anything

Linting each Dockerfile with every pragma stripped shows which rules actually fire. Six do not fire at all under `v2.15.0`:

- `SC2155` in `.docker/cli.dockerfile` - fires on `export`/`local`/`readonly foo=$(...)`; that `RUN` uses a plain `token=$(...)` assignment.
- `DL3018` in `.docker/clamav.dockerfile` - a Debian base image with no `apk`.
- `DL3018` in `.docker/solr.dockerfile` - no package installation in the file.
- `DL3006` in `.docker/database.dockerfile`, `.docker/nginx-drupal.dockerfile` and `.docker/php.dockerfile` - no longer raised for an `ARG`-substituted `FROM`.

A suppression that suppresses nothing is the same blind spot as a global one: it reads as a reviewed exception when nothing is being excepted. These should be removed rather than carried forward with an invented justification.

### Suggested resolution

1. Pin all three invocations to the current release, `hadolint/hadolint:v2.15.0`, and resolve the findings the new rules surface so the pinned version is green rather than merely frozen.
2. Replace every `global ignore` with a per-line `# hadolint ignore=RULE # reason`, and delete the suppressions that no longer fire.
3. Close the Renovate gap. A pin that nothing bumps rots, and the eventual manual bump lands as one large breakage instead of a reviewable PR. Renovate's `github-actions` manager only parses `uses:`, `container:` and `services:`, so an image referenced inside a `run:` block is invisible to it. A `customManagers` regex entry covering `.github/workflows/**` and `.circleci/**` closes that gap, and picks up the existing dclint and gitleaks pins at the same time.
4. Record the check in `.claude/skills/prepare-vortex-release/SKILL.md` so CI tool image versions are verified at release time. With Renovate tracking them this is a verification step rather than a manual bump, but an untagged reference remains a release blocker regardless.

### Related

`#2867` implements all of the above and can close this issue.

Two unrelated unversioned tool invocations found while auditing for the same class of problem are tracked separately, since neither is a hadolint concern.

### Additional observation, out of scope for the fix above

The lint step's loop is fail-fast. Under `bash -e -o pipefail` it aborts at the first failing Dockerfile, so CI reports only the findings from that one file. In this incident the log showed 3 findings in 1 file when there were in fact 8 across 3 files, which would cost three CI round trips to discover by iterating. Collecting a failure flag across the loop and reporting every file before exiting would make the diagnostics match the actual state. Worth considering separately, since it is a diagnostics improvement rather than part of the pinning fix.

AlexSkrypnyk:
Closes #2868

## Summary

Hadolint was invoked as the untagged `hadolint/hadolint`, which resolves to `latest` on every run, while every sibling CI tool image is pinned (`zavoloklom/dclint:3.1.0`, `ghcr.io/gitleaks/gitleaks:v8.30.1`). Hadolint `v2.15.0` shipped on 2026-07-30 with two new rules, so the lint job started failing without a single line of this repository changing.

This pins the image, resolves the findings the new rules raise, replaces every file-level suppression with a justified per-line one, adds a Renovate manager so the pin cannot rot, and records the check in the release skill.

Verified against both releases: `v2.14.0` reports zero findings on all six Dockerfiles, `v2.15.0` reported eight before this change and zero after.

### Pinned the image

`hadolint/hadolint` becomes `hadolint/hadolint:v2.15.0` in `.github/workflows/build-test-deploy.yml`, `.circleci/config.yml`, and `.vortex/tests/lint.dockerfiles.sh`, so both CI providers and a local run all execute the same linter.

### Replaced every global ignore with a per-line one

`# hadolint global ignore=` suppresses a rule for a whole file, including for code added later, so a suppression added for one known line silently blesses every future violation in that file. `DL3064` is a security rule - globally ignored in `cli.dockerfile`, a future `ARG AWS_SECRET_ACCESS_KEY` would pass unnoticed. All five global pragmas are gone; every remaining suppression sits on the line it applies to and carries a short reason, using hadolint's supported `# hadolint ignore=RULE # reason` form so each costs one line:

| File | Rule | Applies to |
| --- | --- | --- |
| `cli.dockerfile` | `DL3064` x3 | `PACKAGE_TOKEN`, `ARG`/`ENV DRUPAL_PRIVATE_FILES` |
| `cli.dockerfile` | `DL3018` | the `apk add` |
| `cli.dockerfile` | `SC2174` | the `mkdir -p -m` |
| `clamav.dockerfile` | `DL3008` | the `apt-get install` |
| `clamav.dockerfile` | `DL3066` x2 | `USER root`, `USER clamav` |
| `database.dockerfile` | `DL3064` | the `MYSQL_*` block |
| `database.dockerfile` | `DL3066` x2 | `USER root`, `USER mysql` |
| `nginx-drupal.dockerfile` | `DL3018` | the `apk add` |
| `php.dockerfile` | `DL3018` | the `apk add` |
| `solr.dockerfile` | `DL3066` x2 | `USER root`, `USER solr` |

None of the findings is a real defect. `PACKAGE_TOKEN` defaults to empty and is supplied at build time via `--mount=type=secret`. `DRUPAL_PRIVATE_FILES` is a directory path matched on the substring `private`. The `MYSQL_*` values are the fixed throwaway credentials the `uselagoon/mysql-8.4` image contract expects for local and CI databases, with deployed environments injecting their own. The `DL3066` hits are the base images' own service accounts, referenced by name because the base images' own tooling refers to them by name. `DL3018` and `DL3008` cover package sets that track the pinned base image.

### Removed six suppressions that suppress nothing

Linting each file with every pragma stripped shows which rules actually fire. Six did not fire at all under `v2.15.0` and were deleted rather than carried forward with an invented justification:

- `SC2155` in `cli.dockerfile` - fires on `export`/`local`/`readonly foo=$(...)`; that `RUN` uses a plain `token=$(...)` assignment.
- `DL3018` in `clamav.dockerfile` - a Debian base image with no `apk`.
- `DL3018` in `solr.dockerfile` - no package installation in the file.
- `DL3006` in `database.dockerfile`, `nginx-drupal.dockerfile` and `php.dockerfile` - no longer raised for an `ARG`-substituted `FROM`.

A suppression that suppresses nothing is the same blind spot as a global one: it reads as a reviewed exception when nothing is being excepted.

### Tracked the pins with Renovate

Added a `customManagers` regex to `renovate.json` covering image references inside `docker run` commands in `.circleci/config.yml` and `.github/workflows/*.yml`. Renovate's built-in `github-actions` manager only parses `uses:`, `container:` and `services:`, and the existing custom manager is scoped to `docker-compose.yml` `IMAGE:` lines, so none of these pins was watched. Validated offline against the real files: the regex matches exactly the seven image references (hadolint, dclint and gitleaks in both providers, plus `rhysd/actionlint` in the test harness) and nothing else - notably not the `"${PWD}":/app` volume arguments or the `${VORTEX_CI_*_IGNORE_FAILURE:-0}` defaults on the same lines. The CI paths were also added to the existing `Container images - All` package rule so these bumps group with the `.docker/**` ones.

### Recorded the check in the release skill

`.claude/skills/prepare-vortex-release/SKILL.md` gains a CI tool image item under Container images. It is a verification step rather than a manual bump, since Renovate now tracks these, and it states that an untagged reference is a release blocker regardless.

### Updated a test assertion

`ToolsHandlerProcessTest` asserted that `# hadolint global ignore=` survives deselecting hadolint. The intent - Dockerfile directives are inert comments that stay useful when the tool is run by hand - is unchanged; the assertion now matches the per-line form.

### Regenerated fixtures

`ahoy update-snapshots` reports 149/149 datasets passing with no fixture drift. Fixtures store the image tag as `hadolint/hadolint:__VERSION__`, so future version bumps produce no fixture churn.

## Screenshots

N/A

## Before / After

```
BEFORE
                                                     
  docker run --rm -i hadolint/hadolint               
                     └─ no tag, resolves to latest   
                                                     
  upstream ships v2.15.0 with DL3064 + DL3066        
                     │                               
                     ▼                               
  8 findings appear overnight, lint job turns red    
  with zero changes in this repository               
                                                     
  5 files carry a file-wide "global ignore",         
  6 of those pragmas suppress nothing at all         
                                                     
  renovate.json watches .docker/** only,             
  so no CI tool image is tracked                     


AFTER
                                                     
  docker run --rm -i hadolint/hadolint:v2.15.0       
                     └─ pinned, same result always   
                                                     
  every suppression sits on the line it applies to   
  and states its reason in one line                  
                     │                               
                     ▼                               
  DL3064 stays active for any future ARG or ENV      
  that has not been individually justified           
                                                     
  renovate.json customManagers regex now watches     
  all 7 CI image pins across both providers          
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

Issue #2889, opened 2026-08-05
Unify CI job identifiers across CircleCI and GitHub Actions


AlexSkrypnyk:
## Summary

Job identifiers in `.circleci/config.yml` and `.github/workflows/` use four different naming styles for the same kind of thing. Unlike step display names, job identifiers surface as status-check contexts, and branch protection rulesets reference those contexts from repository settings that live outside the repository - so renaming a job can silently stop a required check from ever being satisfied.

This was found during a codebase convergence audit of `main`. The step *display names* were converged in that pass; the job *identifiers* were deliberately left alone and raised here, because fixing them safely requires coordinating with the branch ruleset rather than just editing the config.

## Details

The four coexisting styles:

**kebab-case** - `assign-author`, `release-drafter`, `vr-compare`, `update-dependencies`, `vortex-test-common`, `vortex-test-docs`, `vortex-test-installer`.

**snake_case** - `remove_lagoon_environment`.

**Generic placeholder** - `main`, which says nothing about what the job does.

**Bare nouns** - `lint`, `database`, `build`, `deploy`, `audit`.

Related identifier drift in the same files, worth settling in the same pass:

- **Step ids** mix snake_case (`resolve_pr`, `gate`, `compare`, `result`, `summary`) with kebab-case (`check-requirements`) and bare words (`calver`, `version`, `netlify`).
- **Job and step outputs** mix snake_case (`diff_id`, `pr_number`, `changes_percent`, `should_run`, `skipped`) with kebab-case (`release-version`) and a bare `value`.
- **Workflow display names** mix a title-case list (`Database, Build, Test and Deploy`), sentence case (`Security audit`, `Draft release notes`), quoted title case (`Auto Author Assign`, `Close Pull Request`) and a dash-prefixed family (`Test - Visual regression`, `Vortex - Test`).
- Three **filename versus display name** mismatches: `vortex-release.yml` is named `Vortex - Release docs` but also contains the installer job; `vortex-test-common.yml` is named just `Vortex - Test`; `build-test-deploy.yml` is named `Database, Build, Test and Deploy` while its filename omits the database stage.

### Suggested approach

Before renaming any job, enumerate which contexts the branch ruleset currently requires. Rename the job and update the required-context list together, otherwise the protection silently waits forever on a context nothing will ever report.

Step ids, outputs and workflow display names carry none of that risk and can be converged freely in the same change.

Issue #2888, opened 2026-08-05
Unify the public 'VORTEX_*' environment variable vocabulary


AlexSkrypnyk:
## Summary

The `VORTEX_*` environment variables exposed by the shipped tooling scripts in `.vortex/tooling/src` use several different words for the same concept. Because `drevops/vortex-tooling` is a published Composer package and consumer projects set these variables directly in their `.env` files, CI configuration and hosting dashboards, the vocabulary is a public contract - so the inconsistency cannot be fixed on `main` without breaking existing sites.

This was found during a codebase convergence audit of `main`. Every other naming inconsistency surfaced by that audit was converged in place; this one was deliberately deferred here because it needs a deprecation path rather than a rename.

## Details

Roughly 70 variables across 42 scripts are affected. The distinct families:

**Credential nouns - five spellings for one concept.** `_PASS` (FTP, container registry), `_SECRET` (Acquia), `_SECRET_KEY` (S3), `_TOKEN` (GitHub, JIRA, Diffy), `_USER_KEY` (New Relic). The identifier side splits the same way: `_USER`, `_KEY` (Acquia client id), `_ACCESS_KEY` (S3), `_USER_EMAIL` (JIRA).

**Host nouns.** `_HOST` (`VORTEX_FETCH_DB_FTP_HOST`, `VORTEX_FETCH_DB_LAGOON_SSH_HOST`) versus `_HOSTNAME` (`VORTEX_DEPLOY_LAGOON_INSTANCE_HOSTNAME`, `VORTEX_TASK_CUSTOM_LAGOON_INSTANCE_HOSTNAME`).

**Environment nouns.** `_ENVIRONMENT` (`VORTEX_FETCH_DB_ENVIRONMENT`) versus `_ENV` (`VORTEX_TASK_PURGE_CACHE_ACQUIA_ENV`, `VORTEX_NOTIFY_DIFFY_SOURCE_ENV`).

**Lagoon CLI settings.** `LAGOONCLI` fused, with a generic fallback (`VORTEX_DEPLOY_LAGOON_LAGOONCLI_PATH` falling back to `VORTEX_LAGOONCLI_PATH`) versus `CLI` separated, with no fallback (`VORTEX_TASK_CUSTOM_LAGOON_CLI_PATH`).

[...]

PR #2890, opened 2026-08-05
Converged naming, visibility, and idiom conventions across the installer and '.vortex/tests' harness.

### LINKED ISSUE #2925: Installer uses 'private' visibility where the codebase convention is 'protected'
Thirteen members across the installer are declared `private`, while the rest of the codebase uses `protected` throughout.

`private` prevents a subclass from reaching the member at all, which blocks extension and makes the members untestable from a concrete test double. The codebase reserves `public` for the genuine API surface and uses `protected` for everything else; these thirteen are the exception rather than a deliberate choice.

### LINKED ISSUE #2926: Installer method arguments mix camelCase and snake_case
Twenty-seven method arguments across the installer are named in camelCase, while the project convention for local variables and method arguments is snake_case.

This is not purely cosmetic in this codebase: PHPUnit data providers with string keys bind rows to test methods by **named argument**, so a parameter name is part of the contract between a test and its provider. Seventy-eight provider keys are coupled to the affected names and have to move with them. Neither PHPCS, PHPStan nor Rector can see that coupling.

### LINKED ISSUE #2927: Strings use double quotes where no escaping is needed in the installer
Four string literals in the installer use double quotes despite containing no interpolation and no escape sequences, where the project convention is single quotes.

Double quotes signal to a reader that the string is interpolated. Using them for a plain literal makes the reader look for a variable that is not there.

### LINKED ISSUE #2928: Docblocks carry inconsistent '@package' tags in the installer
Thirteen classes carry an `@package` docblock tag while the rest of the installer does not.

`@package` predates PSR-4 and duplicates information the namespace already conveys, so the tag is noise that can and does drift out of step with the actual namespace.

### LINKED ISSUE #2929: 'Validator' boolean predicates are not prefixed to read as questions in the installer
Twenty-two boolean predicate methods on `Validator` are named as bare nouns or verbs rather than as questions, so a call site reads as a statement instead of a test.

A predicate that reads `Validator::gitCommitSha($value)` gives no hint that it returns a boolean, while `Validator::isGitCommitSha($value)` does. The class is a validator, so essentially every public method is a predicate and the convention should be uniform.

Note that some of these predicates call each other internally through `self::`, so the call sites are not all findable by searching for the class name.

### LINKED ISSUE #2930: Installer abbreviates 'destination' as 'dst' in some places and not others
The concept of the install destination is spelled `dst` in some places and `destination` in others, across 251 occurrences in 53 files.

One concept should have one name. `dst` is also close enough to `src` to be misread at a glance.

One constraint applies here: at least one of these constants holds a **public consumer environment variable** as its value. The variable name is part of the installer's public interface and must not change even where the PHP constant does.

### LINKED ISSUE #2931: Subtest helper names in the template test harness do not describe what they are
Helper traits in the template test harness are named as subtests when they contain no subtest methods, across five files.

A trait named for subtests that only provides setup and assertion helpers misleads a reader about where the test steps live.

### LINKED ISSUE #2932: Installer uses 'static::' for references to its own members
One hundred and nine references to a class's own members use `static::` rather than `self::`.

`static::` defers resolution to the runtime class, which is meaningful only where a subclass is expected to override the member. Using it everywhere hides the handful of places where late static binding is genuinely load-bearing, and makes the reader assume an override point exists where none does.

The handler base class is the real exception: it has 42 subclasses and its use of `static::` is deliberate.

### LINKED ISSUE #2933: Null checks and comparison strictness are inconsistent across the installer
Eighteen sites mix `is_null()` with `=== NULL`, and loose with strict comparison, for the same kind of check.

Loose comparison in PHP has surprising edge cases around empty strings, zero and null, so mixing the two forms for equivalent checks makes it unclear at each site whether the looseness was intended.

### LINKED ISSUE #2934: Lazy initialisation is written two different ways in the installer
Two sites initialise a property lazily with an explicit `if ($this->x === NULL)` block, where the rest of the installer uses the null-coalescing assignment operator.

The same idea written two ways for no reason makes a reader stop and look for a difference that is not there.

### LINKED ISSUE #2935: Some files split their 'use' statements into multiple blocks in the installer
Two files separate their `use` statements into more than one block with a blank line between them, breaking the single alphabetically sorted block used everywhere else.

Once a file has two blocks, there is no rule for which block a new import belongs in, so the split tends to grow. PHPCS does not sort namespaces in test files, so nothing catches the drift automatically.

### LINKED ISSUE #2936: Handler lifecycle methods have inconsistent API shapes in the installer
Equivalent lifecycle methods across the prompt handlers differ in signature and return convention, so implementing a new handler means reading several existing ones to work out which shape to follow.

The handlers are the main extension point of the installer, and there are more than forty of them, so a uniform lifecycle contract matters more here than almost anywhere else in the codebase.

### LINKED ISSUE #2937: Installer contains declarations that nothing references
Several declarations in the installer are not referenced anywhere, including from tests.

Unreferenced code still has to be read, understood and maintained, and it misleads anyone trying to work out how a subsystem is wired.

### LINKED ISSUE #2938: 'Archiver' bypasses the project's own file wrapper in the installer
`Archiver` calls the upstream file library directly and creates its temporary directory by hand, while the rest of the installer goes through the project's own `File` wrapper and its temp-directory helper.

The wrapper exists so that path handling and ignore rules are applied consistently. Reaching around it means one component does not get those behaviours.

### LINKED ISSUE #2939: File operation failures throw two different exception types in the installer
Failures from equivalent file operations raise different exception types depending on which code path produced them.

A caller that wants to handle a file failure has to catch more than one type, or catch a base type broad enough to swallow unrelated errors.

### LINKED ISSUE #2940: File deletion is performed several different ways in the installer
File deletion is done through a mix of native calls and the project's own wrapper.

The wrapper accepts either a single path or an array and applies the project's ignore rules, so the native calls silently skip behaviour the rest of the codebase relies on.

### LINKED ISSUE #2941: Some tests are missing the CI group attribute that schedules them in the installer
Several test classes are missing the group attribute used to distribute tests across CI jobs, and attribute ordering is inconsistent where the attributes are present.

A test with no group attribute is not picked up by any CI job, so it silently never runs.

### LINKED ISSUE #2942: Template test harness uses loose comparison for identity checks
Assertions and conditions in the template test harness use loose comparison where identity is intended.

In a test harness this is worse than in production code: a loose comparison can make an assertion pass for a value that is merely coercible to the expected one, so the test reports success without having verified what it claims.

### LINKED ISSUE #2943: Shared assertion traits are composed repeatedly and the project name is hardcoded in the template test harness
Shared assertion traits are composed into several test classes individually rather than once in the common base, and the Docker Compose project name is written out by hand rather than derived from the system under test.

Composing a trait per class means a new test class silently misses assertions that were meant to be universal. A hardcoded project name breaks as soon as the system under test is created under a different directory name.

### LINKED ISSUE #2944: Path joining is written several different ways in the installer
Thirty-six sites across ten files join paths using a mix of the directory separator constant and literal concatenation.

The constant is only meaningful where a path is being **built** for the host platform. Several of these sites use it where a separator is passed as an argument or where the path is normalised anyway, which implies a portability concern that does not exist there and obscures the sites where it genuinely does.

### LINKED ISSUE #2945: 'Config' boolean setters do not match their sibling's fluent return in the installer
Two boolean setters on `Config` return void while their sibling setter returns the instance for chaining.

A caller cannot tell which setters chain without checking each one, so the fluent style is unusable in practice.

AlexSkrypnyk:
Closes #2925
Closes #2926
Closes #2927
Closes #2928
Closes #2929
Closes #2930
Closes #2931
Closes #2932
Closes #2933
Closes #2934
Closes #2935
Closes #2936
Closes #2937
Closes #2938
Closes #2939
Closes #2940
Closes #2941
Closes #2942
Closes #2943
Closes #2944
Closes #2945

## Summary

A convergence pass that removes internal inconsistencies from the Vortex installer and its `.vortex/tests` harness - the kind of drift that accumulates when many contributors and agents touch the same codebase over time.

Eleven Fable readers read every first-party file in full and produced a structured map of divergent naming, API shapes, and idioms; a single Fable planner reviewed the whole map and decided one canonical form per inconsistency class; the user then reviewed and approved every decision across six rounds of questions before anything was applied.

This PR contains 19 commits, one per convergence class, each gated on lint plus the full 1618-test installer suite. Scope is installer internals and the `.vortex/tests` harness only - no shipped template file was touched, so no snapshot regeneration was required and consumer projects are unaffected.

## Changes

Grouped by theme; commits listed oldest to newest within each group.

**Visibility and casing**
- Replaced `private` visibility with `protected` across the installer (`a798bce5`).
- Renamed camelCase method arguments to snake_case in the installer (`87185636`).
- Replaced double-quoted plain strings with single quotes where no escaping is needed (`20d2cb8b`).

**Naming vocabulary**
- Renamed `dst` to `destination` throughout the installer, 251 occurrences across 53 files (`92c476b6`).
- Prefixed `Validator` boolean predicates with `is` (`b8b272ee`).
- Replaced `static::` with `self::` for own-member references (`edccabc9`).
- Aligned subtest helper naming in the template test harness (`79c85c0f`).

**Docblock and import hygiene**
- Removed inconsistent `@package` docblock tags from the installer (`e10c870f`).
- Merged split `use` blocks into a single sorted block (`0334adfa`).

**Comparison and null-check idioms**
- Converged null checks and comparison strictness in the installer (`6631fa27`).
- Converged lazy initialisation on the null-coalescing assignment (`5594f842`).
- Made loose comparisons strict in the template test harness (`cd017c04`).

**Handler lifecycle API shapes**
- Converged handler lifecycle API shapes (`f74fbeb8`): removed lifecycle overrides (`postInstall()`, `hint()`) that only repeated `AbstractHandler`'s own `NULL` default, widened `description()` to its declared `?string` return type across five handlers, and switched `PromptManager` to call `description()` through the class (`$handler::description()`) rather than through an instance.

**Dead-code removal**
- Removed verified-unreferenced declarations from the installer (`e28225aa`).

**File, exception and temp-directory mechanisms**
- Converged `Archiver` on the local `File` wrapper and its temp-dir helper, replacing hand-rolled `sys_get_temp_dir()` / `uniqid()` / `mkdir()` calls with `File::tmpdir()` (`c589540b`).
- Unified the exception type for file failures on `RuntimeException` (`38f2c07b`).
- Converged file deletion on `File::remove()`, replacing `@unlink()` calls and existence-then-remove pairs (`e29085ac`).

**Test-harness composition**
- Added missing CI `#[Group]` attributes and fixed attribute ordering (`d82de6e8`).
- Composed shared assertion traits (`DirectoryAssertionsTrait`, `FileAssertionsTrait`) once on the base functional test case instead of duplicating them across two traits, and derived the Docker Compose project name from the system-under-test path instead of a hardcoded name (`4a4fc050`).

## Not applied

Three decisions from the plan were not carried out; reviewers should see why.

- **`File::exists()` to native `file_exists()`** - the plan's majority was miscounted (actual: 17 files call the native function versus 5 files / 15 call sites using the wrapper). `File::exists()` is the project's own facade alongside `File::remove()`, `File::copy()`, `File::mkdir()`, and `File::dump()`, with a broader `string|array` contract that `file_exists()` does not share. Converging would strip a member out of a deliberate abstraction on a miscounted majority.
- **Typed response getters in `DatabaseImage`, `MigrationImage`, and `MigrationFetchSource`** - their `process()` runs unconditionally via `PromptManager::runProcessors()` even when `->addIf()` skipped the prompt, so `$this->response` is `NULL` on the ordinary path. `getResponseAsString()` throws on a non-scalar value, so converting to it would crash the installer whenever a user picks a database fetch source other than the container registry. The raw `!empty($this->response)` guards are load-bearing, not drift.
- **Converting repeated sibling tests into data providers** - declined by the user during review.

## Remaining work

Roughly 37 further convergence classes were reviewed and approved during the judgment-call rounds but are not part of this PR. Every decision is already made; what is left is applying it. The full specification lives in `.artifacts/tmp/convergence-map/GLOBAL-MAP.md` (the per-class map) and its `PROGRESS.md` ledger.

The main groups still outstanding:

- Remaining installer-internals classes (api-shape and module-pattern): exception types, fluent setters, the Tui output channel, runner invocation, the Logger-aware pattern, composer.json manipulation, path joining, file deletion, temp creation, `Utils\File` in `Archiver`, the queue-vs-execute LAW fix, guard clauses, dead declarations, deduplication, process execution, `dataProvider` placement, dataset keys, test path mirroring, and functional/unit test scaffolding.
- The remaining `.vortex/tests` harness classes: `cmd()` named-tail style, output-assertion style, trait composition, harness drift, `lint.*.sh` alignment, and `#[Group('pN')]` coverage.
- Tooling shell scripts (`drevops/vortex-tooling`): message vocabulary, mechanics, the doc-skeleton rewrite, stray markers, and BATS mechanics/filenames.
- Shipped-template classes, each needing a snapshot regeneration and diff review: em dashes, issue references, final newlines, a dead `.ignorecontent` entry, the `.env` banner, fence tokens, vocabulary, CI step names, the CI safe subset, settings guards, the trusted-host list, scaffold test naming, scaffold extension patterns, demo JS, provision/hook conventions, and the README badge normalizer.
- Four approved behaviour changes: root-ops latent-bug fixes, the timezone-into-CI gap, loosened `=== '1'` env-truthiness checks, and the `your_site_theme` JS behaviour key.

## Related

- #2888 - unify the public `VORTEX_*` environment variable vocabulary.
- #2889 - unify CI job identifiers across CircleCI and GitHub Actions.

Both are filed against milestone 2.0 rather than folded into this PR: they rename consumer-facing contracts and need a deprecation cycle, not an in-place rename.

## Before / After

```
BEFORE - one concept, several spellings               AFTER - one canonical form per class
┌───────────────────────────────────────────┐          ┌─────────────────────────────────────────────┐
│ function download(?string $dst)           │          │ function download(?string $destination)     │
│ private function resolvePromptOverrides() │ ───────► │ protected function resolvePromptOverrides() │
│ static::OPTION_PROFILE                    │          │ self::OPTION_PROFILE                        │
│ Validator::domain() / githubProject()     │          │ Validator::isDomain() / isGithubProject()   │
│ !is_null($discovered)                     │          │ $discovered !== NULL                        │
│ @unlink($f) / exists-then-remove($f)      │          │ File::remove($f)                            │
│ throw new InvalidArgumentException(...)   │          │ throw new RuntimeException(...)              │
└───────────────────────────────────────────┘          └─────────────────────────────────────────────┘
  19 divergent local conventions,                     1 canonical form per class, applied
  scattered across 97 files                            consistently, across the same 97 files
```

### DIFF
--- a/.vortex/installer/src/Command/BuildCommand.php
+++ b/.vortex/installer/src/Command/BuildCommand.php
@@ -30,8 +30,6 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
 
   const OPTION_SKIP_REQUIREMENTS_CHECK = 'skip-requirements-check';
 
-  const TROUBLESHOOTING_URL = 'https://vortex.drevops.com/troubleshooting';
-
   /**
    * Defines default command name.
    *
@@ -117,22 +115,6 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
 
     $this->showFailureSummary();
     return Command::FAILURE;
-  }
-
-  /**
-   * Get the project machine name from .env.
-   */
-  protected function getProjectMachineName(): string {
-    $env_file = $this->destination . '/.env';
-
-    if (file_exists($env_file)) {
-      $content = file_get_contents($env_file);
-      if ($content !== FALSE && preg_match('/^VORTEX_PROJECT=(.+)$/m', $content, $matches)) {
-        return trim($matches[1]);
-      }
-    }
-
-    return basename($this->destination);
   }
 
   /**
--- a/.vortex/installer/src/Command/BuildCommand.php
+++ b/.vortex/installer/src/Command/BuildCommand.php
@@ -47,7 +47,7 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
   /**
    * The working directory for the build.
    */
-  protected string $cwd;
+  protected string $destination;
 
   /**
    * {@inheritdoc}
@@ -68,7 +68,7 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
     Tui::init($output);
 
     $this->isProfile = (bool) $input->getOption(static::OPTION_PROFILE);
-    $this->cwd = $this->getDestination($input);
+    $this->destination = $this->getDestination($input);
 
     if (!$input->getOption(static::OPTION_SKIP_REQUIREMENTS_CHECK)) {
       $requirements_ok = Task::action(
@@ -100,7 +100,7 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
           $env['VORTEX_PROVISION_TYPE'] = 'profile';
         }
 
-        $this->processRunner = $this->getProcessRunner()->setCwd($this->cwd);
+        $this->processRunner = $this->getProcessRunner()->setCwd($this->destination);
         $this->processRunner->run('ahoy build', env: $env);
 
         return $this->processRunner->getExitCode() === RunnerInterface::EXIT_SUCCESS;
@@ -123,7 +123,7 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
    * Get the project machine name from .env.
    */
   protected function getProjectMachineName(): string {
-    $env_file = $this->cwd . '/.env';
+    $env_file = $this->destination . '/.env';
 
     if (file_exists($env_file)) {
       $content = file_get_contents($env_file);
@@ -132,7 +132,7 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
       }
     }
 
-    return basename($this->cwd);
+    return basename($this->destination);
   }
 
   /**
@@ -145,7 +145,7 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
    *   The local development URL, or NULL if it cannot be determined.
    */
   protected function getLocalDevUrl(): ?string {
-    $runner = $this->getProcessRunner()->setCwd($this->cwd);
+    $runner = $this->getProcessRunner()->setCwd($this->destination);
     $runner->run('docker', ['compose', 'config', '--format', 'json'], output: new NullOutput());
 
     if ($runner->getExitCode() === RunnerInterface::EXIT_SUCCESS) {
--- a/.vortex/installer/src/Command/BuildCommand.php
+++ b/.vortex/installer/src/Command/BuildCommand.php
@@ -57,8 +57,8 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
     $this->setDescription('Build the site using ahoy build.');
     $this->setHelp('Checks requirements and runs ahoy build to set up the local site.');
     $this->addDestinationOption();
-    $this->addOption(static::OPTION_PROFILE, 'p', InputOption::VALUE_NONE, 'Build from install profile instead of loading database.');
-    $this->addOption(static::OPTION_SKIP_REQUIREMENTS_CHECK, NULL, InputOption::VALUE_NONE, 'Skip checking for required tools.');
+    $this->addOption(self::OPTION_PROFILE, 'p', InputOption::VALUE_NONE, 'Build from install profile instead of loading database.');
+    $this->addOption(self::OPTION_SKIP_REQUIREMENTS_CHECK, NULL, InputOption::VALUE_NONE, 'Skip checking for required tools.');
   }
 
   /**
@@ -67,10 +67,10 @@ class BuildCommand extends Command implements ProcessRunnerAwareInterface, Comma
   protected function execute(InputInterface $input, OutputInterface $output): int {
     Tui::init($output);
 
-    $this->isProfile = (bool) $input->getOption(static::OPTION_PROFILE);
+    $this->isProfile = (bool) $input->getOption(self::OPTION_PROFILE);
     $this->destination = $this->getDestination($input);
 
-    if (!$input->getOption(static::OPTION_SKIP_REQUIREMENTS_CHECK)) {
+    if (!$input->getOption(self::OPTION_SKIP_REQUIREMENTS_CHECK)) {
       $requirements_ok = Task::action(
         label: 'Checking requirements',
         action: function (): bool {
--- a/.vortex/installer/src/Command/CheckRequirementsCommand.php
+++ b/.vortex/installer/src/Command/CheckRequirementsCommand.php
@@ -79,8 +79,8 @@ class CheckRequirementsCommand extends Command implements ProcessRunnerAwareInte
     $this->setDescription('Check if required tools are installed and running.');
     $this->setHelp('Checks for Docker, Docker Compose, Ahoy, and Pygmy.');
     $this->addDestinationOption();
-    $this->addOption(static::OPTION_ONLY, 'o', InputOption::VALUE_REQUIRED, sprintf('Comma-separated list of requirements to check. Available: %s.', implode(', ', static::REQUIREMENTS)));
-    $this->addOption(static::OPTION_NO_SUMMARY, NULL, InputOption::VALUE_NONE, 'Hide summary with tool versions.');
+    $this->addOption(self::OPTION_ONLY, 'o', InputOption::VALUE_REQUIRED, sprintf('Comma-separated list of requirements to check. Available: %s.', implode(', ', self::REQUIREMENTS)));
+    $this->addOption(self::OPTION_NO_SUMMARY, NULL, InputOption::VALUE_NONE, 'Hide summary with tool versions.');
   }
 
   /**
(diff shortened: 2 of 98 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.55% (204/207)
```
<details>

```
```
</details>

PR #2951, opened 2026-08-11
[#2896] Skipped database image fetch when the image already exists on the host.

### LINKED ISSUE #2896: Database image is pulled from the registry even when it already exists on the host
## Summary

`.vortex/tooling/src/vortex-fetch-db-container-registry` authenticates against the container registry and pulls the database image even when that image is already present on the host. The host is inspected, and the result is used only to print a message - it never influences what the script does next.

The effect is an unnecessary registry login and image pull on every run in that state, costing build time and bandwidth, and replacing whatever image is already on the host. The message printed immediately beforehand, `Found <image> image on host.`, reads as though the local image is about to be used.

### Where the behaviour comes from

The host inspect at line 71 feeds a message and nothing else:

```bash
docker image inspect "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE}" >/dev/null 2>&1 &&
  note "Found ${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE} image on host." ||
  note "Not found ${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE} image on host."
```

Its result is never stored and never tested again.

The only variable gating the fetch is `image_expanded_successfully`. It is initialised to `0` at line 75 and set to `1` in exactly one place - line 85, inside the `if [ -f "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_DB_DIR}/db.tar" ]` branch, after a successful `docker load` and a second successful inspect.

Line 91 then gates the fetch on that variable alone:

```bash
if [ "${image_expanded_successfully}" -eq 0 ]; then
  ...
  "$(dirname "${BASH_SOURCE[0]}")/vortex-login-container-registry"
  docker pull "${VORTEX_FETCH_DB_CONTAINER_REGISTRY}/${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE}"
fi
```

So the fetch is skipped only when an archive was expanded. Image presence on the host, on its own, never skips anything.

### Steps to reproduce

1. Ensure the database image is present on the host, so `docker image inspect <image>` succeeds.
2. Ensure no `db.tar` exists in `VORTEX_FETCH_DB_CONTAINER_REGISTRY_DB_DIR`.
3. Run `vortex-fetch-db-container-registry`.

**Expected:** the image already on the host is used, and no registry call is made.

**Actual:** the script prints `Found <image> image on host.`, then logs in to the registry and pulls the image anyway.

### Why this went unnoticed

The unit test covering this path was named `Skip fetch when image already exists on host`, but it never asserted that a fetch was skipped - it asserted only the `Found ... image on host.` message and overall success. Command mocks were lenient at the time, so the extra `docker login` and `docker pull` calls were answered silently and the test passed.

The stricter mocking that arrived with the `bats-helpers` upgrade rejected those uncovered calls and surfaced the behaviour. In #2894 the test was renamed to `Fetch image when it exists on host and no archive exists` and now asserts the fetch it actually triggers, so the current behaviour is recorded honestly. The script itself was deliberately left unchanged there, because whether it *should* short-circuit is a product decision rather than a test fix.

### Open question

Is the always-pull intentional, to guarantee the freshest database image on every build, or should the script use the host image when it is present and no archive exists?

- **If intentional**, the inspect at line 71 and its `Found ... image on host.` message are misleading and should be reworded to say the image will be refreshed from the registry regardless.
- **If not intentional**, the fetch should be skipped when the inspect succeeds, in the same way an expanded archive skips it.

Either way `.vortex/tooling/tests/unit/fetch-db-container-registry.bats` needs updating alongside, and the renamed test would need inverting if the behaviour changes.

AlexSkrypnyk:
Closes #2896

## Summary
`.vortex/tooling/src/vortex-fetch-db-container-registry` inspected the database image on the host, printed `Found <image> image on host.`, then discarded that result, logged in to the container registry, and pulled the image anyway. The only variable gating the fetch was `image_expanded_successfully`, set exclusively inside the `db.tar` archive branch, so image presence on the host never skipped anything, and every run in that state cost a needless registry authentication and a full multi-hundred-megabyte image pull while replacing whatever image was already on the host.

The fetch is now skipped when the image is already on the host and no archive exists, with `VORTEX_FETCH_DB<index>_FORCE=1` as the override - the same shape the `vortex-fetch-db` router already uses for file-based sources. This makes the documented cache contract true for the `container_registry` source for the first time: `ahoy fetch-db` documents `--fresh` as forcing a fresh database backup and maps it to `VORTEX_FETCH_DB_FRESH=1 VORTEX_FETCH_DB_FORCE=1` (`.ahoy.yml`), and `.vortex/docs/content/development/database.mdx` documents plain `ahoy fetch-db` as using the cache and `--fresh` as bypassing it. Until now `--fresh` was a no-op on this path because the fetch always happened regardless.

Supporting evidence that the always-pull was unintended: the router `vortex-fetch-db` skips its own file-existence cache check only for the `container_registry` source, because the database is an image rather than a file, delegating the caching decision to this script - which then never made it; the archive branch's own comment reads `Always use archived image, even if such image already exists on the host`, which is only meaningful if a host image would otherwise be reused; and `.circleci/vortex-test-common.yml` already sets `VORTEX_FETCH_DB_FORCE: 1` on the job that uses this source, where it was until now a no-op.

Resolution order is now: an expanded `db.tar` archive wins, otherwise an image already on the host is reused, otherwise the image is pulled from the registry with the existing base-image fallback. The archive keeps precedence even when its expansion fails, so that case still fetches rather than silently serving a stale host image.

## Changes
- Added an `archive_file` variable holding the `db.tar` path, replacing four repetitions of the same path expression.
- Converted the host image inspection from an `a && b || c` chain into an explicit `if`/`else` that stores the result in `image_found_on_host`, since the chain could not capture its result without aborting under `set -eu`.
- Read a new `VORTEX_FETCH_DB_FORCE` value from the indexed `VORTEX_FETCH_DB<index>_FORCE` variable, matching the indexing convention the script already uses for its other settings.
- Introduced `should_fetch`, decided once the archive branch has run: `0` when the archive expanded successfully, `0` when the image was found on the host with no archive present and no force flag, `1` otherwise - leaving the existing fetch and base-image-fallback logic unchanged in the branch where it still applies.
- Inverted the unit test that recorded the old always-fetch behaviour and added four tests covering a forced fetch, the indexed `VORTEX_FETCH_DB2_FORCE` variable in both directions, and the failed-archive-expansion fallthrough.
- The skip and fetch tests now assert the exact `docker` call count via `mock_get_call_num` rather than just log output - the original test asserted only the message, which is why the bug survived.

CI impact is nil: `vortex-dev-database-ii`, the only CI job that fetches via the `container_registry` source, already sets `VORTEX_FETCH_DB_FORCE: 1` in `.circleci/vortex-test-common.yml`, and CI build containers start with no local images anyway and consume the database through the cached `.data/db.tar` archive, which retains precedence regardless.

Note, left out of scope so this change stays reviewable against a single issue: `.circleci/vortex-test-common.yml` line 409 sets `VORTEX_FETCH_DB_SOURCE: VORTEX_CONTAINER_REGISTRY`, where every other site uses the lowercase `container_registry` token this router actually matches on - this looks like a separate pre-existing defect.

## Before / After
```
BEFORE ----------------------------------------------------------------
  docker image inspect --> found on host? (logged, then discarded)

  db.tar exists? --yes--> docker load --> expanded ok? --yes--> done
         |no                                    |no
         v                                      v
  ALWAYS --> registry login --> docker pull <----+
  (host image is discarded and replaced, every run)
-------------------------------------------------------------------------

AFTER -----------------------------------------------------------------
  docker image inspect --> image_found_on_host = 0 or 1

  db.tar exists? --yes--> docker load --> expanded ok? --yes--> done
         |no                                    |no
         v                                      v
  image_found_on_host=1 and no archive and FORCE!=1 ?
         |yes --> SKIP fetch, reuse host image
         |no  --> registry login --> docker pull
-------------------------------------------------------------------------
```

## Summary by CodeRabbit

* **New Features**
  * Added an option to force database container image retrieval, even when an image already exists locally.
  * Automatically uses available extracted archives before fetching from the registry.
  * Skips unnecessary registry downloads when a suitable local image is available.

* **Bug Fixes**
  * Falls back to registry retrieval when archive extraction fails.
  * Improved handling for indexed database configurations.

### DIFF
--- a/.vortex/tooling/src/vortex-fetch-db-container-registry
+++ b/.vortex/tooling/src/vortex-fetch-db-container-registry
@@ -45,6 +45,10 @@ _v="VORTEX_FETCH_DB${_db_index}_CONTAINER_REGISTRY_IMAGE_BASE"
 _vs="VORTEX_DB${_db_index}_IMAGE_BASE"
 VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE_BASE="${!_v:-${!_vs:-}}"
 
+# Force fetch even if the image already exists on the host.
+_v="VORTEX_FETCH_DB${_db_index}_FORCE"
+VORTEX_FETCH_DB_FORCE="${!_v:-}"
+
 #-------------------------------------------------------------------------------
 
 # @formatter:off
@@ -68,15 +72,21 @@ info "Started database data container image fetch."
 [ -z "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_PASS}" ] && fail "Missing required value for VORTEX_FETCH_DB_CONTAINER_REGISTRY_PASS or VORTEX_CONTAINER_REGISTRY_PASS." && exit 1
 [ -z "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE}" ] && fail "Destination image name is not specified. Please provide VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE or VORTEX_DB_IMAGE in a format <org>/<repository>." && exit 1
 
-docker image inspect "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE}" >/dev/null 2>&1 &&
-  note "Found ${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE} image on host." ||
+archive_file="${VORTEX_FETCH_DB_CONTAINER_REGISTRY_DB_DIR}/db.tar"
+
+image_found_on_host=0
+if docker image inspect "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE}" >/dev/null 2>&1; then
+  note "Found ${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE} image on host."
+  image_found_on_host=1
+else
   note "Not found ${VORTEX_FETCH_DB_CONTAINER_REGISTRY_IMAGE} image on host."
+fi
 
 image_expanded_successfully=0
-if [ -f "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_DB_DIR}/db.tar" ]; then
-  task "Found archived database container image file ${VORTEX_FETCH_DB_CONTAINER_REGISTRY_DB_DIR}/db.tar. Expanding..."
+if [ -f "${archive_file}" ]; then
+  task "Found archived database container image file ${archive_file}. Expanding..."
   # Always use archived image, even if such image already exists on the host.
-  docker load -q --input "${VORTEX_FETCH_DB_CONTAINER_REGISTRY_DB_DIR}/db.tar"
+  docker load -q --input "${archive_file}"
 
   # Check that image was expanded and now exists on the host or notify
   # that it will be downloaded from the registry.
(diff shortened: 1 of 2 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #2971, opened 2026-08-12
[#2967] Moved 'exit 1' into the 'fail()' helper across the shell tooling.

### LINKED ISSUE #2967: Move `exit 1` to `fail()` in tooling

AlexSkrypnyk:
Closes #2967

## Summary

Every Vortex shell script declared its own single-line `fail()` helper that only printed a `[FAIL] ...` message, leaving termination to the call site, so `exit 1` was hand-written roughly 230 times across the 49 scripts that define it: the 42 shipped tooling scripts in `.vortex/tooling/src/`, the 6 `scripts/*.sh` template scripts, and the documented boilerplate example. `fail()` now ends with `exit "${2:-1}"`, so a single call both prints the message and terminates the script, defaulting to status 1 with an optional second argument for a caller-supplied status, and every redundant `exit 1` that used to follow a `fail()` call has been removed. A handful of failure blocks were also reordered so `fail()` always runs last, and three failure paths that previously exited silently with no `[FAIL]` line now route through it.

## Changes

- Added `exit "${2:-1}"` to the end of `fail()` in all 42 shipped tooling scripts under `.vortex/tooling/src/`, the 6 `scripts/*.sh` template scripts, and the documented boilerplate at `.vortex/docs/content/contributing/maintenance/script-boilerplate.sh` - including the differently-formatted `fail() { printf "     ! %s\n" "${1}"; }` variant used by the `provision-*.sh` subscripts.
- Removed the now-redundant `&& exit 1` suffix from roughly 160 inline `<condition> && fail "..." && exit 1` guards; under `set -e` a false condition was never the last command of the AND-OR list either before or after this change, so failure behaviour is unchanged.
- Collapsed 26 `for cmd in ...; do command -v "${cmd}" >/dev/null || { fail "..."; exit 1; }; done` command-availability guards from four lines to one.
- Removed dozens of standalone `exit 1` lines that immediately followed a `fail()` call.
- Reordered several failure blocks so `fail()` runs last: explanatory `note` lines, cleanup `rm` calls, payload dumps, and `drush` diagnostics now all run before the terminating `fail()`, so nothing that used to print is now unreachable. Affects `vortex-doctor`, `vortex-import-db-file`, `vortex-provision`, `vortex-fetch-db-url`, `vortex-notify-jira`, `vortex-notify-github`, and `scripts/provision-20-migration.sh` (and its 9 installer-fixture copies).
- In `vortex-notify-github`, a second consecutive `fail()` call that was really an advisory follow-up line - a workaround for `fail()` previously not being terminal - is now a `note`, printed before the real `fail()` call.
- Routed three failure paths that previously exited with no `[FAIL]` line through `fail()`: `vortex-provision` (a corrupted container-image database), `vortex-push-db-s3` (a raw `echo "ERROR: ..."` is now `fail "..."`, with the message text unchanged apart from the prefix), and `scripts/provision-20-migration.sh` (a silent `exit 1` after a `drush migrate:messages` dump now gets a `[FAIL]` headline).
- `vortex-deploy-lagoon` captured a deployment's exit status and re-raised it as the script's own exit code in a final check, and `scripts/vortex-tooling.sh`'s `composer_run()` helper returned a captured status for `set -e` to propagate immediately; both now pass that status to `fail()` as its second argument instead, so the exit code in either case is preserved exactly.
- Updated `.vortex/docs/content/contributing/maintenance/template.mdx` (rules 6-8) and its companion `script-boilerplate.sh` example to the new helper and idiom, including fixing the boilerplate's `command -v curl >/dev/null || ( fail "..." && exit 1 )` subshell form - whose `exit` only ever killed the subshell, never the script - to a plain `|| fail "..."`.
- Updated `.vortex/tooling/tests/unit/push-db-s3.bats` for the `ERROR: ` to `[FAIL] ` prefix change, and regenerated 14 installer test fixtures via `ahoy update-snapshots` to match the simplified scripts they embed.

## Screenshots

N/A

## Before / After

```
BEFORE - fail() only prints; every call site repeats the termination ───────

  fail() { ...; printf "[FAIL] %s\n" "${1}"; }

  [ -z "${VAR}" ] && fail "Missing VAR." && exit 1

  for cmd in curl; do command -v "${cmd}" >/dev/null || {
    fail "Command ${cmd} is not available."
    exit 1
  }; done

  if [ ! -f "${dump_file}" ]; then
    fail "Unable to import database from file."     # [FAIL] printed here
    note "Dump file ${dump_file} does not exist."
    exit 1                                           # context printed after
  fi

AFTER - fail() prints and terminates; call sites stop repeating it ─────────

  fail() { ...; printf "[FAIL] %s\n" "${1}"; exit "${2:-1}"; }

  [ -z "${VAR}" ] && fail "Missing VAR."

  for cmd in curl; do command -v "${cmd}" >/dev/null || fail "Command ${cmd} is not available."; done

  if [ ! -f "${dump_file}" ]; then
    note "Dump file ${dump_file} does not exist."    # context printed first
    fail "Unable to import database from file."      # [FAIL] + exit, last
  fi
```

### DIFF
--- a/.vortex/docs/content/contributing/maintenance/script-boilerplate.sh
+++ b/.vortex/docs/content/contributing/maintenance/script-boilerplate.sh
@@ -18,13 +18,13 @@ VORTEX_EXAMPLE_URL="${VORTEX_EXAMPLE_URL:-http://example.com}"
 info() { [ "${TERM:-}" != "dumb" ] && tput colors >/dev/null 2>&1 && printf "\033[34m[INFO] %s\033[0m\n" "${1}" || printf "[INFO] %s\n" "${1}"; }
 note() { printf "       %s\n" "${1}"; }
 pass() { [ "${TERM:-}" != "dumb" ] && tput colors >/dev/null 2>&1 && printf "\033[32m[ OK ] %s\033[0m\n" "${1}" || printf "[ OK ] %s\n" "${1}"; }
-fail() { [ "${TERM:-}" != "dumb" ] && tput colors >/dev/null 2>&1 && printf "\033[31m[FAIL] %s\033[0m\n" "${1}" || printf "[FAIL] %s\n" "${1}"; }
+fail() { [ "${TERM:-}" != "dumb" ] && tput colors >/dev/null 2>&1 && printf "\033[31m[FAIL] %s\033[0m\n" "${1}" || printf "[FAIL] %s\n" "${1}"; exit 1; }
 # @formatter:on
 
 info "Started Vortex operations."
 
-[ -z "${VORTEX_EXAMPLE_URL}" ] && fail "Missing required value for VORTEX_EXAMPLE_URL." && exit 1
-command -v curl >/dev/null || (fail "curl command is not available." && exit 1)
+[ -z "${VORTEX_EXAMPLE_URL}" ] && fail "Missing required value for VORTEX_EXAMPLE_URL."
+command -v curl >/dev/null || fail "curl command is not available."
 
 # Example of the script body.
 curl -L -s -o /dev/null -w "%{http_code}" "${VORTEX_EXAMPLE_URL}" | grep -q '200\|403' && note "Requested example page."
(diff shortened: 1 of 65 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #2973, opened 2026-08-12
Forward-ported 13 commits from 'main' onto '2.x'.

AlexSkrypnyk:
## Summary

Forward-ports 13 commits from `main` onto `2.x`. `main` has kept moving while `2.x` restructured three subsystems - the installer relocated to `.vortex/cli`, the tooling scripts were rewritten from Bash to PHP, the shell provision scripts were replaced by `drupal/deploy_steps`, and `drevops/ci-runner` was dropped in favour of the CI providers' base images - so most of what landed on `main` either already exists here or no longer has a place to land. All 72 undecided `main` commits were analysed against the current state of `2.x` before anything was applied; 13 were ported, 59 were deliberately excluded.

Ten commits were clean version bumps or self-contained fixes. Three needed adapting because the code they target has been rewritten on this branch.

### Security audit workflow (adapted)

`main` split Gitleaks and `composer audit` out of the lint job into a standalone `Security audit` workflow so a failing audit is distinguishable from a failing linter and can be re-run on its own. `2.x` still ran both inline, so the split applies - but the source commit's new job runs in the `drevops/ci-runner` container that `2.x` removed. The ported `audit.yml` runs on `ubuntu-latest` with `shivammathur/setup-php`, matching how every other job on this branch is built, and drops the container-executor workarounds (`DOCKER_CONFIG`, the `$HOME` preservation step) that only existed to service that image.

The equivalent `audit` job and workflow were added to `.circleci/config.yml`, `CiProvider.php` now removes `audit.yml` when GitHub Actions is deselected, and the documentation was updated. Two documentation conflicts were resolved against this branch's job set rather than `main`'s: `2.x` has no nightly `database` job, so the CI pipeline diagram and the CircleCI job list keep their own shape.

### Database image fetch skip (adapted)

Re-implemented against this branch's PHP `vortex-fetch-db-container-registry` rather than the Bash script the source commit patched. When the image is already on the host and there is no archive to expand, the fetch is skipped; `VORTEX_FETCH_DB{N}_FORCE=1` restores the unconditional fetch. An expanded archive still wins over the host image, preserving the existing precedence. Three PHPUnit cases cover the new branch points, including the archive-beats-host-image case.

### `fail()` helper (adapted)

`fail()` now exits by itself, so failure branches no longer append `&& exit 1`, and it accepts an optional exit status. Narrowed to the two places that still ship shell on this branch - `scripts/vortex-tooling.sh` and the maintenance script boilerplate - since the tooling scripts the source commit also touched are PHP here.

### CircleCI deploy branch filter

The single 189-character deploy regex is split into a list of per-pattern regexes, each with its own comment, and `CircleCiConfigTest` gains a `testFilterPatternLength` case asserting every filter pattern in the config stays within CircleCI's 128-character limit. This branch's `getCommitWorkflowJob()` name-based job lookup was kept over the source commit's hardcoded `jobs[3]` index, so the assertion stays correct as jobs move around.

### Dependency and version bumps

- Drupal core `~11.4.4` → `~11.4.5`
- `sdc_devel` `^1.0.3`, `drupal-rector` `^1.1.2`, `phpstan` `^2.2.8`, `rector` `^2.6.1`
- Lagoon cli/nginx/php images `26.7.0` → `26.8.0`, `clamav-debian` `1.5.3` → `1.5.4`
- `selenium/standalone-chromium` `150.0` → `151.0` (kept this branch's vendor-agnostic `browser` service name)
- `hadolint` `v2.15.1`, `zizmor-action` `v0.6.2`, `renovatebot/github-action` `v46.2.2`
- CI database cache key `v26.7.2` → `v26.8.0`, applied as a token replacement because `main` still carries a separate `database` job that `2.x` folded into `test`

### Deliberately excluded

59 commits were analysed and not ported. Most are already effectively present here under a different commit. The ones worth calling out because a reviewer might expect them:

- **Three convergence sweeps** (`44fce96`, `5a14578`, `8da57df` - the largest touching 320 files) are stylistic naming, structure and comment sweeps whose payload lands almost entirely in `.vortex/installer` and the Bash tooling. Replaying `main`'s patches is not meaningful here; the equivalent work is to re-run that convergence against this branch's own tree, which is worth a separate pass.
- `0a9ffdc` (`VORTEX_DB_DIR`/`VORTEX_DB_FILE` ignored for indexed fetches) - the bug does not exist here. The PHP rewrite's `getenv_default()` fallback chains already include the unindexed variables.
- `10b840e` (anchored environment name matching) - exact by construction here: `RebuildSearchIndex.php` uses `in_array(..., TRUE)`, and the shell provision scripts it patched no longer exist.
- `17640a4` (`vortex-tooling.sh` aborting on an older package directory) - unreachable here, because the early-exit guard already removes the stale copy before the `mv`.
- `b04386a` (docs workflows building the v2 binary from `.vortex/installer`) - a `main`-only fix. This branch's workflows are already correct: they build themselves from `.vortex/cli` and the other major from `.vortex/installer`.
- `aa246e3` (`drevops/ci-runner` bump), `e94fde1` (provision script split), `ebe2519` and `7175b12` (BATS tests) - all target subsystems that no longer exist here.
- `03039ec` (`.vortex` dev dependency refresh) - touches `.vortex/installer/composer.{json,lock}` and diverged lock files; this branch refreshes its own.
- `8718f34` (1.40.0 release prep) - `main`-only; pins tooling to `1.3.0` where this branch requires `^2.0@alpha`.

## Snapshots

`ahoy update-snapshots` was run and the regenerated fixtures are committed in `Updated snapshots.` (124 files). The new `audit.yml` appears in `_baseline`, with `-audit.yml` removal markers across the CircleCI scenarios and a gitleaks-stripped variant in `gitleaks_disabled` - matching the fixture shape the source commit produced on `main`. `.circleci/vortex-test-common.yml` was regenerated via `php .vortex/tests/generate-vortex-dev-circleci`, and the lint suite's freshness check confirms it is up to date.

No template file was deleted by any applied commit, so no `SutTrait.php` follow-up is outstanding.

## Gates

Local, all passing:

- `ahoy lint` from `.vortex/` - phpcs, phpstan, rector, shellcheck, hadolint, markdownlint, cspell, and the generated-CircleCI-config freshness check
- `actionlint` 1.7.12 with the same ignore set CI uses - clean, including the new `audit.yml`
- CLI test suite - 1685 tests, 5379 assertions
- Tooling test suite - 806 tests, 3837 assertions
- Template `CircleCiConfigTest` - 177 tests, including the 26 new filter-pattern-length datasets that confirm every pattern is within CircleCI's limit

CI runs the full template-test workflow, which is the final source of truth.

## Before / After

```
Security audit: inline in the lint job  ──►  its own workflow

  BEFORE                                  AFTER

  ┌─ lint ─────────────────────┐          ┌─ lint ─────────────────────┐
  │ composer validate          │          │ composer validate          │
  │ hadolint / dclint          │          │ hadolint / dclint          │
  │ gitleaks          ◄────────┼──┐       │ build stack                │
  │ build stack                │  │       │ install dev deps           │
  │ install dev deps           │  │       │ composer normalize         │
  │ composer audit    ◄────────┼──┤       │ phpcs / phpstan / rector   │
  │ composer normalize         │  │       └────────────────────────────┘
  │ phpcs / phpstan / rector   │  │
  └────────────────────────────┘  │       ┌─ audit (own workflow) ─────┐
                                  └──────►│ gitleaks                   │
  A failing audit looked like a           │ composer audit --locked    │
  failing linter, and re-running          └────────────────────────────┘
  it meant re-running everything.
                                          Runs on ubuntu-latest + setup-php
                                          (not the removed ci-runner image).
                                          Re-runnable on its own.
```

```
Database image fetch: host image reuse

  BEFORE                                  AFTER

  image on host?                          image on host?
       │                                       │
       ├─ yes ─┐                               ├─ yes ─┐
       └─ no  ─┤                               └─ no  ─┤
               ▼                                       ▼
        archive exists?                         archive exists?
               │                                       │
        ┌──────┴──────┐                         ┌──────┴──────┐
        ▼             ▼                         ▼             ▼
      expand      pull from                   expand    on host and
        │          registry                     │       not forced?
        ▼             │                         │             │
      done ◄──────────┘                         │       ┌─────┴─────┐
                                                │       ▼           ▼
  The host image was always                     │   reuse it   pull from
  discarded and re-pulled, even                 ▼                registry
  when it was already correct.                done ◄──────────────┘

                                          VORTEX_FETCH_DB{N}_FORCE=1
                                          restores the old behaviour.
```

```
CircleCI deploy filter: one regex  ──►  a list of patterns

  BEFORE                                  AFTER

  only: /^(production|main|master|        only:
  develop)$|^project\/[a-zA-Z0-9\-\.]       # production, main, master, develop
  +|^(feature|bugfix)\/[a-zA-Z0-9\-\.\      - /^(production|main|master|develop)$/
  ,_]+$|^ci.*|^(release|hotfix)\/[0-9]      # project/description
  +(\.[0-9]+){2}(-rc\.[0-9]+)?$|^(rel       - /^project\/[a-zA-Z0-9\-\.]+/
  ease|hotfix)\/[0-9]{4}-[0-9]{2}-[0-       # feature/... and bugfix/...
  9]{2}(\.[0-9]+)?$|^[0-9]+\.x$/            - /^(feature|bugfix)\/[a-zA-Z0-9...]+$/
                                            # ci, cisomething
  189 chars in one line, past                - /^ci.*/
  CircleCI's 128-char limit, with            ... one pattern per intent
  comments detached above it.
                                          Each pattern carries its own comment,
                                          and a test now asserts none exceeds
                                          128 characters.
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (153/153)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (153/153)
```
<details>

```
```
</details>

PR #2998, opened 2026-08-13
[#2996] Published the shipped tooling script variables in the generated docs table.

### LINKED ISSUE #2996: Generated variables docs omit every variable declared in the shipped tooling scripts
## Summary

The generated variables table in `.vortex/docs/content/development/variables.mdx` contains no variable declared in the shipped tooling scripts under `.vortex/tooling/src/`. The table currently has 137 rows, all sourced from `.env`, `.env.local.example`, `scripts/`, and the `extra/` stubs, while the tooling scripts declare roughly 295 distinct `VORTEX_*` and `DRUPAL_*` variables between them. None of those are published.

The `Vortex - Test docs` gate cannot catch this. `.github/workflows/vortex-test-docs.yml` regenerates the table and fails on any diff, but the regeneration produces nothing for those paths either, so the gate passes on an omission rather than reporting it. The same is true of the equivalent step in `.github/workflows/vortex-release.yml`.

## Details

`.vortex/docs/.utils/update-docs.sh` passes `.vortex/tooling/src` to `drevops/shellvar:1.7.0 extract` as one of its input paths. Shellvar's directory traversal finds nothing there, because the shipped scripts are extensionless executables (`vortex-export-db-file`, `vortex-provision`, `vortex-notify-slack`, and so on) rather than `*.sh` files. Handed an explicit file path, shellvar parses the same file correctly.

Reproduce from the repository root:

```bash
# Directory form - returns an empty table.
docker run --rm -v "$PWD:/app" drevops/shellvar:1.7.0 extract --exclude-local --format=md-table /app/.vortex/tooling/src

# Explicit file form - extracts the variables, their defaults and their descriptions.
docker run --rm -v "$PWD:/app" drevops/shellvar:1.7.0 extract --exclude-local --format=md-table /app/.vortex/tooling/src/vortex-export-db-file
```

The second command prints both `VORTEX_EXPORT_DB_FILE_DIR` and `VORTEX_EXPORT_DB_FILE_STRUCTURE_TABLES` with the correct default values and descriptions, confirming the parser handles these files and only the directory walk skips them.

Points to settle:

- Whether to expand the directory to explicit file paths inside `update-docs.sh`, or to raise it upstream in shellvar so its traversal also picks up extensionless files carrying a shell shebang. The upstream fix helps every consumer; the local fix lands sooner and is a few lines.
- Whether the docs gate needs a positive assertion in addition to the diff check. A diff check alone cannot distinguish "documentation is current" from "the generator produced nothing", which is exactly the failure here. Asserting that a known tooling variable appears in the output would close that gap.
- Expect a large one-off diff when this is fixed, since roughly 295 variables enter the table at once. Worth landing on its own rather than inside an unrelated change.

Noticed while adding `VORTEX_EXPORT_DB_FILE_STRUCTURE_TABLES` in #2984: the new variable never reached the published table, and the docs gate stayed green.

AlexSkrypnyk:
Closes #2996

## Summary

The docs variables table at `.vortex/docs/content/development/variables.mdx` is generated by handing five input paths to `drevops/shellvar:1.7.0 extract`. One of those inputs, the directory `.vortex/tooling/src`, contributed zero rows: shellvar's directory traversal only descends into `*.sh` and `*.bash` files, and every one of the 41 shipped tooling scripts there (`vortex-provision`, `vortex-notify-slack`, and the rest) is an extensionless executable, so no tooling variable was ever published. `update-docs.sh` now expands that directory into explicit file paths, which shellvar parses correctly, and gained two guards so a similar omission fails the build instead of silently producing a table that looks current. Publishing the tooling variables also surfaced two source-comment accuracy issues, both fixed here, plus one internal variable pair worth excluding from the table.

## Changes

- `.vortex/docs/.utils/update-docs.sh` - expands the `.vortex/tooling/src` directory into an array of explicit file paths before handing them to shellvar, since shellvar only walks `*.sh`/`*.bash` files and the tooling scripts are extensionless; shellvar now extracts their variables instead of silently contributing nothing.
- `.vortex/docs/.utils/update-docs.sh` - adds a glob-empty guard (fails if the tooling directory expansion matches no files) and an `assert_source_present` check that each of the five input sources is actually represented in the extracted table before it is written. This closes a gap in the `Vortex - Test docs` CI gate: that gate regenerates the table and fails on any `git diff`, but a diff check alone cannot distinguish "documentation is current" from "the generator produced nothing for this input" - exactly how the tooling omission stayed green. Both assertions run against the raw table, before the `sed` passes alias the `extra/` stub paths onto display labels and before the output file is touched, so a failed generation leaves `variables.mdx` unmodified. The assertion strings use literal backticks, which needed a `shellcheck disable=SC2016` alongside the file's existing `SC2129` disable to keep `ahoy lint-scripts` clean.
- `.vortex/docs/content/development/variables.mdx` - regenerated; the table grows from 137 to 385 variables.
- `.vortex/docs/.utils/variables/variables.excluded.txt` - adds `SCRIPT_DIR` and `TARGET_ENV_REMAP`. Both are internal script values (a script-directory path and a loop-local remap) rather than configuration knobs, joining the same class of already-excluded tooling internals (`SRC_TMPDIR`, `ENV_ID`, `VORTEX_PROVISION_LOG_ACTIVE`, `VORTEX_DOCTOR_CHECK_PREFLIGHT`).
- `.vortex/tooling/src/vortex-notify` - wraps the `<name>.log` placeholder in backticks, since MDX parsed the bare angle brackets as a JSX tag and broke the Docusaurus build; also moves `VORTEX_NOTIFY_LOG_FILE`'s comment to describe the value it holds rather than the log-collection block that fills it, with that block's own rationale moved down beside the block itself.
- `.vortex/tooling/src/vortex-deploy-artifact` - corrects the swapped descriptions of `VORTEX_DEPLOY_ARTIFACT_GIT_USER_NAME` and `VORTEX_DEPLOY_ARTIFACT_GIT_USER_EMAIL`; the name variable was documented as an email address and vice versa.

## Screenshots

Variables page showing tooling script variables

## Before / After

```
┌──────────────────────────────────────────────────────────────────────────────┐
│ BEFORE                                                                       │
├──────────────────────────────────────────────────────────────────────────────┤
│ update-docs.sh hands shellvar one directory as input:                        │
│                                                                              │
│   .env  .env.local.example  .vortex/tooling/src  scripts  extra/             │
│                                                                              │
│ shellvar walks a directory input, matching only *.sh / *.bash names.         │
│ vortex-provision, vortex-notify-slack, ... are extensionless                 │
│   -> directory input matches nothing -> 0 rows contributed                   │
│                                                                              │
│ No check that an input actually produced rows.                               │
│                                                                              │
│ variables.mdx: 137 variables, 0 from the 41 shipped tooling scripts.         │
│ CI diff-check: green (output never changed, so there is nothing to diff).    │
└──────────────────────────────────────────────────────────────────────────────┘
                                       │
                                       ▼
┌──────────────────────────────────────────────────────────────────────────────┐
│ AFTER                                                                        │
├──────────────────────────────────────────────────────────────────────────────┤
│ update-docs.sh expands the directory into explicit file paths first:         │
│                                                                              │
│   tooling_paths=(.vortex/tooling/src/vortex-provision ...41 files)           │
│   [ "${#tooling_paths[@]}" -gt 0 ] || exit 1   (glob-empty guard)            │
│                                                                              │
│   .env  .env.local.example  "${tooling_paths[@]}"  scripts  extra/           │
│                                                                              │
│ shellvar parses each file explicitly -> tooling variables extracted.         │
│                                                                              │
│ assert_source_present checks all 5 input sources landed in the table,        │
│ before the extra/ path-aliasing sed and before the file is written           │
│   -> a future zero-yield source fails the build, not a silent diff.          │
│                                                                              │
│ variables.mdx: 385 variables, tooling rows present and asserted.             │
└──────────────────────────────────────────────────────────────────────────────┘
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #3008, opened 2026-08-14
[#3007] Closed every 'task' with a 'pass' or a 'fail' in shipped scripts.

### LINKED ISSUE #3007: Close every 'task' with 'pass' or 'fail' in shipped scripts
## Summary

`.vortex/CLAUDE.md` now states that every `task` in a shipped script must be closed by a `pass` or a `fail`, so a step that announces itself always reports its outcome. The rule was added in #3006, but only the two `task` calls introduced by that PR were brought in line with it.

A sweep of the shipped scripts finds **55 `task` calls across 19 files** that announce work and never report an outcome. The reader sees `> Retrieving authentication token.` and then nothing, with no way to tell whether the step succeeded, was skipped, or silently failed.

## Details

Counts per file:

| File | Unclosed `task` calls | Lines |
|------|----------------------|-------|
| `.vortex/tooling/src/vortex-deploy-lagoon` | 8 | 154, 194, 222, 262, 283, 291, 319, 359 |
| `.vortex/tooling/src/vortex-task-purge-cache-acquia` | 6 | 83, 88, 93, 98, 135, 160 |
| `.vortex/tooling/src/vortex-export-db-image` | 5 | 51, 56, 59, 62, 74 |
| `.vortex/tooling/src/vortex-task-copy-db-acquia` | 5 | 87, 92, 97, 102, 107 |
| `.vortex/tooling/src/vortex-task-copy-files-acquia` | 5 | 83, 88, 93, 98, 103 |
| `.vortex/tooling/src/vortex-setup-ssh` | 4 | 105, 112, 114, 126 |
| `.vortex/tooling/src/vortex-notify-jira` | 3 | 147, 152, 169 |
| `.vortex/tooling/src/vortex-reset` | 3 | 45, 52, 55 |
| `.vortex/tooling/src/vortex-deploy-artifact` | 2 | 118, 120 |
| `.vortex/tooling/src/vortex-fetch-db-lagoon` | 2 | 117, 132 |
| `.vortex/tooling/src/vortex-notify-newrelic` | 2 | 150, 160 |
| `.vortex/tooling/src/vortex-push-container-registry` | 2 | 88, 101 |
| `.vortex/tooling/src/vortex-task-custom-lagoon` | 2 | 75, 91 |
| `.vortex/tooling/src/vortex-fetch-db-acquia` | 1 | 245 |
| `.vortex/tooling/src/vortex-fetch-db-container-registry` | 1 | 84 |
| `.vortex/tooling/src/vortex-login-container-registry` | 1 | 60 |
| `.vortex/tooling/src/vortex-provision` | 1 | 351 |
| `scripts/provision-10-enable-dev-modules.sh` | 1 | 55 |
| `scripts/provision-40-example.sh` | 1 | 46 |

The two entries under `scripts/` ship into consumer projects as provision subscripts, so they set the example that consumers copy - worth fixing first.

### The rule

From `.vortex/CLAUDE.md`, Tooling package / Script pattern:

> Every `task` MUST be closed by a `pass` or a `fail`. A task announces work that is starting, so it always reports its outcome; a `task` with no closing line leaves the reader unable to tell whether the step succeeded. This holds even when the work itself cannot fail (e.g. a command suffixed with `|| true`) - close it with `pass`.

### How the list was produced

For each script, walk the lines and track an open `task`. A `pass` or `fail` closes it; reaching the next `task`, `info` or `note`, or the end of the file, with one still open is a finding. Roots scanned: `scripts/`, `.vortex/tooling/src/`, `hooks/library/` (`hooks/library/` was clean).

The heuristic is line-based, so a handful of entries may turn out to be intentional - for example a `task` whose work genuinely continues past an intervening `note`. Each one needs a look rather than a blind fix.

### Scope

- Add the missing `pass` (or `fail`) line to each site, wording the outcome as the past-tense counterpart of the task (`Retrieving authentication token.` closes with `Retrieved authentication token.`).
- Update the BATS output assertions in `.vortex/tooling/tests/unit/` for every affected script, since the new lines change script output.
- Re-run `ahoy update-snapshots` from `.vortex/` for the `scripts/` changes, which are copied into the installer fixtures.

Not in scope: changing what any script does, or reworking the `info`/`note`/`task` helpers themselves.

Follow-up to #3006.

AlexSkrypnyk:
Closes #3007

## Summary

`.vortex/CLAUDE.md` requires every `task` in a shipped script to be closed by a `pass` or a `fail`, so a step that announces itself always reports its outcome. This sweeps the shipped scripts and closes every `task` that did not report one, so a reader can always tell whether a step succeeded, was skipped, or failed.

The scope was re-derived rather than taken from the issue list. The issue's counts came from a line-based heuristic that treats an intervening `note` as a terminator, which both over- and under-reports. Re-running the scan with `note` kept subordinate to the open `task`, then adjudicating every site by hand, gives **61 sites across 18 files** instead of the reported 55 across 19.

This also fixes a defect the issue does not mention: `task()` records `_TASK_START` and `pass()` prints the elapsed time and unsets it, so an unclosed `task` leaked its start time into the next `pass`, which then reported a duration measured from the wrong step.

### Shipped tooling scripts

Added the missing closing `pass` to 16 scripts under `.vortex/tooling/src/`, wording each as the past-tense counterpart of its task.

Two scripts announced their *success* with a `note` and never emitted a `pass` at all: `vortex-task-copy-db-acquia` and `vortex-task-copy-files-acquia` now close with `pass "Copied database from ... environment."` and `pass "Copied files from ... environment."`.

### Provision subscripts

`scripts/provision-00-enable-demo-modules.sh` and `scripts/provision-20-migration.sh` each had an `else` branch that ended in a `note`, leaving the task open on that path. Both branches now close with a `pass`, following the precedent already in `vortex-notify-jira` and `vortex-provision`, where a skip or a fallback is reported as a real outcome.

### Tests

Added assertions locking in the new closers across 13 BATS files. The existing assertions are unordered substring checks, so nothing broke on its own; the new assertions are what actually guard the rule. Verified they have teeth by deliberately corrupting one expected string and confirming the suite failed.

Three `deploy-lagoon` failure-path tests assert the *absence* of a success line, which is the property that matters: a failed deploy must never print `[ OK ]`.

### Fixtures

Regenerated via `ahoy update-snapshots`: 11 datasets, covering `custom_modules_none` for the demo-modules subscript and the nine `migration_*` datasets for the migration subscript.

## Where the scope differs from the issue

Two patterns defeat a naive scan and account for nearly all the difference.

A `pass` or `fail` guarded by `[ ... ] &&` closes only the guard's path, not the task's main path, as in `vortex-notify-jira` where `[ -z "${issue}" ] && pass ... && exit 0` leaves the found-issue path open. A `pass` that is the first statement after an `else` belongs to the sibling branch rather than to the task above the `if`, as in `vortex-fetch-db-acquia` where the cached-file branch's `pass` was mistaken for the fetch task's closer.

Roughly 14 of the issue's entries are already closed today and need no change. Three files drop out entirely: `vortex-provision:351`, `scripts/provision-10-enable-dev-modules.sh:55` and `scripts/provision-40-example.sh:46`. `vortex-deploy-lagoon` drops from 8 sites to 4.

Roughly 20 real sites were not in the issue list, including `vortex-deploy-lagoon:380`, seven further sites in `vortex-fetch-db-acquia`, `vortex-push-container-registry:88`, `vortex-reset:58`, `vortex-setup-ssh:141`, and the two `else`-branch sites under `scripts/`.

## Judgment calls

**Lagoon deploys close conditionally.** `run_lagoon_deploy` always returns 0 and records the outcome in `exit_code`; the script prints the CLI error and calls `fail` once at the end. An unconditional `pass` would print `[ OK ] Deployed environment` immediately before `[FAIL] Lagoon deployment completed with errors`, defeating the rule it is meant to satisfy. Failing at the call site is also wrong, because two of the four sites are followed by override-flag restore or removal and exiting there would leave a stray `VORTEX_PROVISION_OVERRIDE_DB` on the environment. The four sites therefore use `[ "${exit_code}" = "0" ] && pass ...`, matching the file's existing one-liner idiom, and failure stays with the deferred end-of-script `fail`.

**Wording follows the task, except where that would overstate.** A Lagoon deploy is *requested* rather than done, because the CLI queues a build and the script then sleeps waiting for the queue. An Acquia backup is *requested* for the same reason, since the very next task waits for it to complete.

**Nested tasks keep their nesting.** Three sites open a `task` inside another. Both are closed rather than restructuring the scripts, which the issue puts out of scope. `pass()` already handles an unset `_TASK_START`, so the outer line simply prints without a duration. Demoting the inner `task` to a `note` would be cleaner output and is worth a follow-up.

**Two sites were deliberately left alone.** `vortex-fetch-db-acquia:175` and `vortex-setup-ssh:80` already end every path in a `pass` or a `fail`.

**One site cannot satisfy the rule literally.** In `vortex-task-purge-cache-acquia`, a domain that cannot be purged is deliberately non-fatal so the remaining domains still run, and `fail` exits. That path keeps its `Warning:` note and gains a neutral `pass "Completed cache purging for ..."`, so the task is closed either way without a failure being labelled `[ OK ]`.

## Before / After

```text
BEFORE                                   AFTER
──────                                   ─────

[TASK] Retrieving authentication token.   [TASK] Retrieving authentication token.
[TASK] Retrieving app UUID.               [ OK ] Retrieved authentication token. (1s)
[TASK] Retrieving environment ID.         [TASK] Retrieving app UUID.
[ OK ] Finished fetch. (43s)              [ OK ] Retrieved app UUID. (1s)
                                          [TASK] Retrieving environment ID.
   ▲                                      [ OK ] Retrieved environment ID. (1s)
   │                                      [ OK ] Finished fetch. (2s)
   └── every step announced, none
       reported; the final duration           ▲
       is measured from the FIRST             │
       unclosed task, not from the            └── every step reports its own
       step it is printed against                 outcome, and each duration
                                                  is measured from its own task


 task/closer pairing                      task/closer pairing
 ───────────────────                      ───────────────────

   task ─┐                                  task ─┐
         │  (no closer)                           │
   task ─┤                                  pass ─┘   _TASK_START consumed
         │                                  task ─┐
   task ─┤                                        │
         │                                  pass ─┘
   pass ─┘  ◀── closes the wrong task       task ─┐
            _TASK_START still held                │
            from the first task            pass ─┘
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #3018, opened 2026-08-15
[#3017] Installed 'testmode' from the development modules provision script.

### LINKED ISSUE #3017: Install every selected dev-only module in 'provision-10-enable-dev-modules.sh'
## Summary

`scripts/provision-10-enable-dev-modules.sh` installs `sdc_devel`, `devel` and `generated_content`, but the installer's `modules` catalogue also offers `testmode`, which has no other install path. A project that selects Testmode gets `drupal/testmode` in `composer.json` and nothing that ever installs it, so any Behat scenario tagged `@testmode` fails.

Testmode is test-only and is deliberately excluded from the exported configuration, so provisioning is the only place it can be installed. That makes the omission silent: the package is present, the module is not, and the failure surfaces later as a Behat failure rather than as a provisioning error.

## Details

The script currently installs three modules:

```bash
task "Installing Single Directory Component development tools."
drush pm:install sdc_devel || true

task "Installing Devel module."
drush pm:install devel || true

task "Installing Generated content module."
GENERATED_CONTENT_CREATE=1 drush pm:install generated_content
```

The `modules` prompt in `--schema` offers eighteen options. Most are site modules that belong in exported config, but the dev/test-only ones need an install step here because they are intentionally kept out of `core.extension.yml`.

Expected: a project that selects a dev/test-only module from the catalogue has that module installed after provisioning.

Actual: only `sdc_devel`, `devel` and `generated_content` are installed. `testmode` is selectable but never installed.

### Question

Should this script install every dev/test-only module the project selected, rather than a fixed set? If the list is meant to stay fixed, it would help to document which catalogue entries are expected to be installed by the consumer project rather than by the template, so the gap is visible at selection time instead of at test time.

### Context

Found while updating a consumer project from `1.40.2` to `eaceb9ed110b5dceaa2752acf4faf7fb4606f8f3`. The project had its own version of this script that installed `testmode`; adopting the template's version dropped it, and the regression was only caught by comparing the two scripts by hand.

AlexSkrypnyk:
Closes #3017

## Summary

`testmode` is one of the 18 modules in the installer's catalogue, but nothing in the template installed it directly - the only install path was transitive, through `web/modules/custom/ys_demo/ys_demo.info.yml` declaring `- testmode:testmode`, which pulled the module in whenever `drush pm:install ys_demo` ran in `scripts/provision-00-enable-demo-modules.sh`. A project that keeps Testmode selected but drops the demo module - the normal outcome once a project stops relying on demo content - never installs Testmode, so any `@testmode`-tagged Behat scenario fails at runtime when `TestmodeTrait` calls `Testmode::getInstance()` on a module that was never enabled. `settings.testmode.php` also puts Testmode in `config_exclude_modules`, so configuration import can never enable it either, leaving no path to recover once the demo module is gone.

This PR installs Testmode directly from `scripts/provision-10-enable-dev-modules.sh`, alongside Devel and SDC Devel, and fixes the corresponding deselection path so removing Testmode through the installer no longer leaves the demo module hard-depending on a package that was just removed.

## Changes

- **Provisioning script** - `scripts/provision-10-enable-dev-modules.sh` gets a new fenced `MODULE_TESTMODE` block that runs `drush pm:install testmode`, with no `|| true` (unlike the neighbouring `devel` and `sdc_devel` lines), so a failed install aborts provisioning immediately instead of resurfacing later as a Behat failure.
- **Installer selection handling** - `.vortex/installer/src/Prompts/Handlers/Modules.php` adds `testmode` to `DEV_MODULES` so the provisioning script survives when Testmode is the only development module selected, and rewrites the docblock to state the actual criterion for that list. It also deletes `tests/behat/features/pages.feature` when Testmode is deselected - that feature's only scenario asserts Testmode filtering but is not named after the module, so the installer's generic `<module>.feature` removal previously missed it.
- **Deselection coupling fix** - fenced `MODULE_TESTMODE` blocks now wrap the `testmode:testmode` dependency in `ys_demo.info.yml`, the `use Drupal\testmode\Testmode;` import together with the `ys_demo_deploy_configure_testmode()` deploy hook in `ys_demo.deploy.php`, and the `TestmodeTrait` import/use in `tests/behat/bootstrap/FeatureContext.php` (mirroring the existing `MODULE_REDIRECT` fences in the same file). Before this change, deselecting Testmode removed `drupal/testmode` from `composer.json` while the demo module kept a hard dependency on it, so `drush pm:install ys_demo` would fail during provisioning. This half of the change sits adjacent to the issue title but is included deliberately - it is the other side of decoupling Testmode from the demo module, and shipping only the install half would leave a selection combination (Testmode deselected, demo module kept) that could never provision.
- **Deploy file layout** - the Testmode import and the deploy hook that uses it are grouped into a single fenced block placed immediately after the imports in `ys_demo.deploy.php`, and the fence markers carry `phpcs:disable` / `phpcs:enable` for `Squiz.WhiteSpace.FunctionSpacing`. `File::removeToken()` is purely line-based and strips only the marker lines, so a blank line added inside a fence leaks into installed projects while one left outside it survives deselection. That leaves no marker placement around a standalone top-level function that satisfies the blank-line sniffs in all three states at once - the template source with markers, the installed project with Testmode selected, and the installed project with Testmode deselected. Grouping the import with its hook puts the opening marker before a `use` statement rather than after a closing brace, which is the one position where the surrounding sniff still measures a clean gap. All three states were rendered and linted to confirm.
- **Tests** - BATS coverage added to `.vortex/tooling/tests/unit/provision-enable-dev-modules.bats` (all 4 scenarios) and `.vortex/tooling/tests/unit/provision.bats` (13 scenarios) to assert the new install step. Installer functional scenarios in `ModulesHandlerProcessTest.php` strengthened: `modules_no_testmode` now asserts both the demo-module coupling and `pages.feature` are gone, `modules_no_devel_sdc_devel_generated_content` now asserts the script survives carrying only the Testmode install, and a new `modules_no_devel_sdc_devel_generated_content_testmode` scenario asserts the script itself is removed once all four development modules are deselected.
- **Docs** - `.vortex/docs/content/drupal/modules.mdx`, `provision.mdx` and `testmode.mdx` updated so the documented install path matches the shipped behaviour; `modules.mdx` now states explicitly that `seckit` is the only catalogue entry the template deliberately does not install.
- **Fixtures** - installer fixtures under `.vortex/installer/tests/Fixtures/handler_process/` regenerated to match the handler and script changes, via `ahoy update-snapshots`.

## Before / After

```
BEFORE - testmode reachable only through the demo module
┌───────────────────────────────────────────┐
│ ys_demo.info.yml                          │
│   dependencies:                           │
│     - testmode:testmode                   │
└─────────────────────┬─────────────────────┘
                      │ drush pm:install ys_demo
                      ▼
          testmode installed ONLY IF ys_demo is kept

  demo module removed, Testmode kept selected:
          testmode never installed
          -> @testmode Behat scenarios fail
          -> config_exclude_modules blocks config import too

  Testmode deselected, demo module kept:
          ys_demo.info.yml still requires testmode:testmode
          -> drush pm:install ys_demo fails on a removed package

AFTER - testmode installed directly, decoupled from the demo module
┌───────────────────────────────────────────┐
│ provision-10-enable-dev-modules.sh        │
│   drush pm:install sdc_devel || true      │
│   drush pm:install devel || true          │
│   #;< MODULE_TESTMODE                     │
│   drush pm:install testmode               │
│   #;> MODULE_TESTMODE                     │
└─────────────────────┬─────────────────────┘
                      │ runs whenever Testmode is selected
                      ▼
          testmode installed independent of the demo module

  demo module removed, Testmode kept selected:
          testmode installs directly from provisioning

  Testmode deselected, demo module kept:
          the #;< MODULE_TESTMODE fences remove the
          dependency, the import and the deploy hook
          together with the module
          -> drush pm:install ys_demo succeeds
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #3037, opened 2026-08-19
[#3035] Added 'CI_BEHAT_PROFILE_OFFSET' to decouple the Behat profile from the runner index.

### LINKED ISSUE #3035: Add 'CI_BEHAT_PROFILE_OFFSET' so the Behat profile does not depend on the runner index
## Summary

The Behat profile name is derived from the runner index, so `p0` is only reachable when Behat runs on runner 0. Because `p0` is the catch-all, any project that dedicates runner 0 to a once-only tool loses every untagged scenario - and the job still passes green.

Adding a `CI_BEHAT_PROFILE_OFFSET` variable, defaulting to `0`, decouples the profile number from the index without renaming a single profile or re-tagging a single feature file.

### Current behaviour

Both providers derive the profile name from the runner index:

```bash
# .github/workflows/build-test-deploy.yml:607
# .circleci/config.yml:551 (and .circleci/vortex-test-common.yml)
if [ "${CI_RUNNER_TOTAL}" -gt 1 ]; then export VORTEX_CI_BEHAT_PROFILE="${VORTEX_CI_BEHAT_PROFILE:-p${CI_RUNNER_INDEX}}"; fi
```

The profile name *is* the index. `behat.yml` defines `p0` as the catch-all:

```yaml
# Runs all tests tagged with '@smoke' or not tagged with '@p1', and not tagged
# with '@skipped'. This is a 'catch-all' profile that runs any tests not tagged
# with '@pX'.
p0:
  gherkin:
    filters:
      tags: '@smoke,~@p1&&~@skipped'
```

### Why this is a problem

The docs already tell consumers that a tool given its own container must take the *last* container, never the first, precisely because container 0 owns `p0`. That guidance is correct, but it is guidance rather than a mechanism, and the failure mode when it is not followed is silent.

If Behat stops running on runner 0, `p0` is never selected by anybody. Every scenario carrying no `@pN` tag stops running, and nothing fails - no profile is missing, no filter errors, the job goes green. A missing `pN` fails loudly; an unselected catch-all does not fail at all.

Today the only way for a project to free runner 0 is to renumber the profiles in `behat.yml` and re-tag every feature file, which is a large mechanical change to test code in service of a CI layout decision.

### Proposed change

Introduce `CI_BEHAT_PROFILE_OFFSET`, subtracted from the index before it becomes a profile name:

```bash
if [ "${CI_RUNNER_TOTAL}" -gt 1 ]; then export VORTEX_CI_BEHAT_PROFILE="${VORTEX_CI_BEHAT_PROFILE:-p$((CI_RUNNER_INDEX - CI_BEHAT_PROFILE_OFFSET))}"; fi
```

Declared alongside the existing runner-role flags:

```yaml
# GitHub Actions - job env
CI_BEHAT_PROFILE_OFFSET: 0
```

```bash
# CircleCI - Set test runner roles
echo "export CI_BEHAT_PROFILE_OFFSET=0"
```

The profiles then stay numbered from `p0` regardless of which runner Behat starts on. A project pinning the once-only tools to runner 0 sets the offset to `1` and gets:

| Runner | Role | Profile |
|--------|------|---------|
| 0 | PHPUnit, coverage, SDC validation | none |
| 1 | Behat | `p0` (catch-all) |
| 2 | Behat | `p1` |

`behat.yml` is untouched and no feature file is re-tagged.

### Why the default stays inert

With `CI_BEHAT_PROFILE_OFFSET: 0` the expression reduces to `p${CI_RUNNER_INDEX}`, which is exactly today's behaviour. The generated template behaves identically, and consumers opt in with a single integer. The single-runner case is unaffected either way, because the `-gt 1` guard short-circuits before the subtraction and the profile stays `default`.

### Limitation

The offset only works while the Behat runners are one contiguous block starting at the offset. Behat on runners 0 and 2 but not 1 is not expressible. That was judged an acceptable trade for the small diff and the identical mechanism on both providers.

### Files affected

- `.github/workflows/build-test-deploy.yml` - declare the variable in the job `env:` block, use it in the Behat step.
- `.circleci/config.yml` and `.circleci/vortex-test-common.yml` - declare it in the `Set test runner roles` step, use it in the Behat step.
- `behat.yml` - the profile comment currently asserts that "'pN' requires a runner N", which the offset makes untrue. It should describe the offset instead.
- `.vortex/docs/content/continuous-integration/README.mdx` - the parallelism section, including the callout stating that a dedicated tool container must be the last one. With the offset that constraint is lifted, and the docs should say so.
- `.vortex/installer/tests/Fixtures/handler_process/**` - regenerate with `ahoy update-snapshots`.

### Alternatives considered

- **Behat runner list** (`CI_BEHAT_RUNNERS="1,2"`, position in the list becomes the profile number). Handles non-contiguous sets and collapses two declarations into one, but a job-level `env:` entry in GitHub Actions cannot read another job-level `env:` entry, so the list has to be duplicated in the `CI_IS_BEHAT_RUNNER` condition or moved to a repository variable. CircleCI has no such problem, so the two providers stop mirroring each other.
- **Explicit profile map** (`CI_BEHAT_PROFILE_MAP='-,p0,p1'`, one field per runner). The most expressive - it frees profile names from the `pN` convention entirely - but the map length has to track the runner count by hand, and a short map yields an empty field, so it needs a mandatory guard. Worth reconsidering if freeing the naming convention turns out to matter more than diff size.
- **`strategy.matrix.include`** carrying a `behat_profile` per entry. Fully declarative and the smallest change on GitHub Actions, but CircleCI's `parallelism: N` yields N identical nodes with no equivalent, so it would undo the provider symmetry.

### Worth pairing with this

None of the above fixes the underlying hazard on its own: a catch-all that is never selected still passes green. A cheap independent backstop is to fail the Behat step when the run matched zero scenarios, which also catches a typo in a tag and a filter that quietly stopped matching.

### Prior art

The offset mechanism is in use on a consumer project with a three-instance matrix, pinning PHPUnit, coverage and SDC validation to instance 0 and running Behat on instances 1 and 2 with the stock `p0` and `p1` profiles.

AlexSkrypnyk:
Closes #3035

## Summary

Both CI providers derived the Behat profile name directly from the runner index (`p${CI_RUNNER_INDEX}`), and `p0` is the catch-all profile that runs every untagged scenario, so a project that dedicated runner 0 to once-only tools and moved Behat to later runners never selected `p0` and its untagged scenarios silently stopped running while CI stayed green. This change adds `CI_BEHAT_PROFILE_OFFSET`, a value subtracted from the runner index before it becomes a profile number (`p$((CI_RUNNER_INDEX - CI_BEHAT_PROFILE_OFFSET))`), declared alongside the existing `CI_IS_*_RUNNER` role flags. The default of `0` keeps behaviour byte-identical to before, so existing projects are unaffected. A project that frees leading runners for other tools sets the offset once and Behat runners map back to `p0`, `p1`, and so on, with no changes to `behat.yml` or feature file tags.

## Changes

- Declared `CI_BEHAT_PROFILE_OFFSET` in the GitHub Actions test job's `env:` block and used it in the Behat step's profile derivation.
- Exported `CI_BEHAT_PROFILE_OFFSET` in the CircleCI `Set test runner roles` step and used it in the Behat step, mirrored across `.circleci/config.yml` (consumer config) and `.circleci/vortex-test-common.yml` (the template's own test harness).
- Rewrote the `behat.yml` profiles header comment to describe the derived profile number and clarify that only runners which run Behat need a profile.
- Reworked the parallelism documentation page with a worked example (once-only tools pinned to runner 0, Behat on runners 1-2 mapped to `p0`/`p1` via offset `1`, with a role table), a note that `behat.yml` stays untouched, and a warning scoping the contiguity limitation to runs without an explicit `VORTEX_CI_BEHAT_PROFILE`.
- Added a note to the Behat development page linking profile numbering to the offset and the parallelism page.
- Regenerated the installer test fixtures under `.vortex/installer/tests/Fixtures/handler_process/` to match the CI config and `behat.yml` changes (unified-diff snapshots, tooling-generated).

## Before / After

```
BEFORE: profile hardwired to the runner index
+------------------------------------------------------------+
| profile = p${CI_RUNNER_INDEX}                              |
|                                                              |
| Runner 0 -> p0   (must run Behat, or p0 silently skipped)  |
| Runner 1 -> p1                                              |
| Runner 2 -> p2                                              |
+------------------------------------------------------------+

AFTER: profile derived via an offset, default 0 (identical to BEFORE)
+------------------------------------------------------------+
| profile = p$((CI_RUNNER_INDEX - CI_BEHAT_PROFILE_OFFSET))  |
|                                                              |
| CI_BEHAT_PROFILE_OFFSET=1 (opt-in, frees runner 0):         |
|   Runner 0 -> once-only tools (Jest, PHPUnit, SDC)          |
|   Runner 1 -> p0   (catch-all)                              |
|   Runner 2 -> p1                                            |
+------------------------------------------------------------+
```

## Summary by CodeRabbit

- **New Features**
  - Parallel Behat test runs now support leading CI containers that perform other tasks, while preserving consistent profile numbering.

- **Documentation**
  - Added guidance and examples for configuring profile offsets across supported CI systems.
  - Clarified profile numbering, container roles, validation warnings, and setup requirements for additional runners.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -390,8 +390,13 @@ jobs:
               #;> MODULE_SDC_DEVEL
               #;> DRUPAL_THEME
               #;< TOOL_BEHAT
-              # Runs on every node, using the `p<node index>` profile.
+              # Runs on every node, using the `p<node index - offset>` profile.
               echo "export CI_IS_BEHAT_RUNNER=1"
+              # Subtracted from the node index to derive the profile number.
+              # Raise it by one for every leading node that does not run
+              # Behat, so the first Behat node still selects the `p0`
+              # catch-all profile.
+              echo "export CI_BEHAT_PROFILE_OFFSET=0"
               #;> TOOL_BEHAT
             } >> "${BASH_ENV}"
 
@@ -548,7 +553,7 @@ jobs:
           name: Test with Behat
           command: |
             [ "${CI_IS_BEHAT_RUNNER:-1}" = "1" ] || exit 0
-            if [ "${CI_RUNNER_TOTAL:-1}" -gt 1 ]; then export VORTEX_CI_BEHAT_PROFILE="${VORTEX_CI_BEHAT_PROFILE:-p${CI_RUNNER_INDEX}}"; fi
+            if [ "${CI_RUNNER_TOTAL:-1}" -gt 1 ]; then export VORTEX_CI_BEHAT_PROFILE="${VORTEX_CI_BEHAT_PROFILE:-p$((CI_RUNNER_INDEX - CI_BEHAT_PROFILE_OFFSET))}"; fi
             echo "Running with ${VORTEX_CI_BEHAT_PROFILE:-default} profile"
             docker compose exec -T -e BEHAT_SCREENSHOT_ANIMATION_SKIP=1 cli php -d memory_limit=-1 vendor/bin/behat --colors --strict --profile="${VORTEX_CI_BEHAT_PROFILE:-default}" || \
               docker compose exec -T -e BEHAT_SCREENSHOT_ANIMATION_SKIP=1 cli php -d memory_limit=-1 vendor/bin/behat --colors --strict --rerun --profile="${VORTEX_CI_BEHAT_PROFILE:-default}" || \
--- a/.circleci/vortex-test-common.yml
+++ b/.circleci/vortex-test-common.yml
@@ -225,8 +225,13 @@ jobs:
               #;> MODULE_SDC_DEVEL
               #;> DRUPAL_THEME
               #;< TOOL_BEHAT
-              # Runs on every node, using the `p<node index>` profile.
+              # Runs on every node, using the `p<node index - offset>` profile.
               echo "export CI_IS_BEHAT_RUNNER=1"
+              # Subtracted from the node index to derive the profile number.
+              # Raise it by one for every leading node that does not run
+              # Behat, so the first Behat node still selects the `p0`
+              # catch-all profile.
+              echo "export CI_BEHAT_PROFILE_OFFSET=0"
               #;> TOOL_BEHAT
             } >> "${BASH_ENV}"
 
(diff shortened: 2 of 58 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #3062, opened 2026-08-21
[#3060] Synchronised npm lock files when the installer removes front-end tools.

### LINKED ISSUE #3060: Regenerate the lock file when the installer removes front-end tool dependencies
## Summary

When the installer removes a front-end tool, it edits `package.json` but leaves `package-lock.json` untouched. The two then disagree, and `npm ci` refuses to install: it treats the lock file as authoritative and aborts rather than reconciling it.

This affects a generated project on its first build, so it surfaces as a broken site rather than as a warning.

## Details

The gap is partial deselection. When every front-end tool is deselected, the `frontend_all` group removes `package.json` and `package-lock.json` together, so that path is consistent. When only some are deselected - ESLint off with Jest kept, for example - `package.json` is rewritten through `JsonManipulator::updateFile()`, the lock file is left as it was, and the pair no longer matches.

The behaviour is not specific to ESLint. Every tool in the map edits `package.json` the same way, so any combination that removes one tool and keeps another reaches it. It also predates the move to npm: `yarn.lock` was committed before that change and `yarn install --frozen-lockfile` fails on the same mismatch, so only the error message changed.

- [ ] Keep `package.json` and its lock file consistent whenever the installer removes tool dependencies, for every tool in the map rather than for one of them.
- [ ] Cover partial deselection, where one tool is removed and another is kept, since full deselection already removes both files together.
- [ ] Apply the same treatment to the custom theme, which carries its own manifest and lock file.
- [ ] Add installer coverage for a partial deselection, asserting the resulting manifest and lock file install cleanly.

Found during review of #3047: https://github.com/drevops/vortex/pull/3047#discussion_r3819590674

AlexSkrypnyk:
Closes #3060

## Problem

When the installer removed a front-end tool it rewrote `package.json` via `JsonManipulator::updateFile()` but left `package-lock.json` untouched. `ahoy fei` runs `npm ci`, which treats the lock as authoritative and aborts when the pair disagrees, so a generated project failed on its first build. Full deselection was safe because the `frontend_all` group removes both files together; partial deselection (ESLint off, Jest kept) was the broken path.

## Fix

A new `DrevOps\VortexInstaller\Utils\NpmLock` utility reconciles the lock file with the manifest next to it. `NpmLock::sync()` copies the manifest's dependency blocks (`dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`) onto the lock's root `packages[""]` entry, then garbage-collects every `packages` entry no longer reachable from a root, following node's resolution rules: the innermost `node_modules` copy wins, falling back outwards to the hoisted tree. It is offline and deterministic, with no npm invocation, because Node is not an installer prerequisite. It writes the lock the way npm does: two-space indent, unescaped slashes, trailing newline.

It is wired in at the manifest edit itself, in the new `Tools::updateNpmManifest()`, so the sync is derived from the tool map's single list of package names rather than a second list that could drift out of sync. Every tool in the map gets it, not just the one that exposed the bug.

## Custom theme treatment

The custom theme carries its own `package.json` and `package-lock.json`, with the full ESLint, Prettier and Stylelint dependency sets and its own lint scripts, and the tool map never touched them, so a project that deselected ESLint still installed and ran ESLint inside the theme, and `ahoy lint-fe` calls the theme's `npm run lint` directly. New `theme.package.json` entries on the ESLint and Stylelint tool definitions clean the theme manifest, and its lock is synced the same way as the root pair. The theme manifest and lock are never deleted, even when every front-end tool is deselected, because they also drive the Sass/PostCSS/Terser build.

## Dead aggregate scripts

With both linters deselected, each linter's callback rewrote `lint` to call the other one, leaving `lint` and `lint-fix` pointing at scripts that no longer exist. The `frontend_linting` group now removes that pair from both manifests, drops the theme lint lines from `.ahoy.yml`, and drops the two matching rows from the theme `README.md` command table.

## CI configs

The theme lint step in `.github/workflows/build-test-deploy.yml` and `.circleci/config.yml` was fenced only on `DRUPAL_THEME`, so with both linters deselected CI still ran `npm run lint` in the theme against a script that had been removed. Both steps now sit inside the `TOOL_ESLINT_STYLELINT` fence as well.

## Refactor

The content-removal pass (`strings` and `lines`) was extracted from `processTool()` into `processContent()` and is now shared with `processGroup()`, so groups can declare content removals the same way tools do. It is skipped when a definition declares neither key.

## Tests

New `NpmLockTest` unit suite (18 cases) covers root reconciliation, orphan pruning, retention of transitives shared with a surviving dependency, nested `node_modules` shadowing and fall-back to the hoisted tree, workspace/link entries as roots, unresolvable optional dependencies, a missing lock, an unsupported lock, non-object dependency blocks, npm-shaped output formatting and empty-object round-tripping. `ToolsHandlerProcessTest` gains manifest/lock sync assertions on the partial-deselection scenarios plus a new `tools_no_eslint_no_stylelint` fixture scenario (both linters off, Jest kept). Snapshots regenerated.

Fixture snapshots continue to ignore lock file content via `.ignorecontent`, so the lock is asserted programmatically in the tests rather than as an 8,600-line fixture diff.

## Verification

The PHP prune was checked against real npm 10.9.2: on the template's own root pair it produced exactly the same 421-entry package set as `npm install --package-lock-only`, with no differences beyond `libc` metadata that npm drops and this preserves. Running the installer end-to-end with both linters deselected then produced a project where `npm ci --dry-run` succeeds at the root (331 packages) and in the theme (175 packages).

## Screenshots

N/A

## Before / After

```
BEFORE - partial deselection (ESLint off, Jest kept)

    package.json                     package-lock.json
    -------------------------        -------------------------
    "devDependencies": {             "packages": {
      "jest": "^29.0.0"                "": {
    }                                    "devDependencies": {
                                            "eslint": "^9.0.0",
                                            "jest": "^29.0.0"
                                          }
                                        },
                                        "node_modules/eslint": { ... },
                                        "node_modules/jest": { ... }
                                      }

    $ npm ci
    npm error `npm ci` can only install packages when your package.json and
    npm error package-lock.json are in sync. Please update your lock file.
    npm error Missing: eslint@9.0.0 from lock file
    npm error
    npm error Build fails on the first run.

AFTER - NpmLock::sync() reconciles the pair

    package.json                     package-lock.json
    -------------------------        -------------------------
    "devDependencies": {             "packages": {
      "jest": "^29.0.0"                "": {
    }                                    "devDependencies": {
                                            "jest": "^29.0.0"
                                          }
                                        },
                                        "node_modules/jest": { ... }
                                        (eslint and its orphans pruned)
                                      }

    $ npm ci
    added 331 packages in 4s
    Build succeeds.
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

PR #3084, opened 2026-09-03
[#3078] Kept 'auth.json' out of the container build context and image.

### LINKED ISSUE #3078: Keep 'auth.json' out of the container build context and image
## Summary

`.dockerignore` does not exclude `auth*`, and `.docker/cli.dockerfile` copies it explicitly. A project with a local `auth.json` therefore bakes its package token into the built image.

The `--mount=type=secret,id=package_token` mount already supplies that token to the Composer install, so the file copy is redundant as well as unsafe.

## Details

`.docker/cli.dockerfile` at 1.41.0:

```dockerfile
COPY composer.json composer.* patches.lock.* .env* auth* /app/

RUN --mount=type=secret,id=package_token \
    token=$(if [ -s /run/secrets/package_token ]; then cat /run/secrets/package_token; else echo "${PACKAGE_TOKEN}"; fi) && \
    if [ -n "${token}" ]; then export COMPOSER_AUTH="{\"github-oauth\": {\"github.com\": \"${token}\"}}"; fi && \
    COMPOSER_MEMORY_LIMIT=-1 composer install -n --no-dev --ansi --prefer-dist --optimize-autoloader
```

Two paths put the file in the image: the explicit `COPY ... auth* /app/` above, and the later `COPY . /app`, because the 1.41.0 deny-list `.dockerignore` does not list `auth*`.

Credentials for the install come from `COMPOSER_AUTH`, built from the build secret or the `PACKAGE_TOKEN` build arg - never from `auth.json`. So nothing in the build needs the file.

This is not a regression introduced by the deny-list rewrite: the previous allow-list `.dockerignore` carried an explicit `!auth.json`, so the file was already entering the build context. The rewrite preserved the behaviour rather than causing it.

### Suggested change

Add the credentials file to `.dockerignore`:

```
# Credentials. The package token reaches the build through a build secret.
auth*
```

and drop it from the copy:

```dockerfile
COPY composer.json composer.* patches.lock.* .env* /app/
```

Verified on a consumer project: the CLI image builds and Composer authenticates unchanged with both applied.

Worth considering alongside this: `auth.json` is not in the template's `.gitignore` either, so a token can also reach the repository.

### How this surfaced

Raised by an automated review on a consumer project's 1.41.0 update PR, then verified against the template.

AlexSkrypnyk:
Closes #3078

## Summary

`.dockerignore` did not exclude `auth*`, and `.docker/cli.dockerfile` copied it explicitly, so a project with a local `auth.json` baked its Composer package token into the built image through two paths: the explicit `COPY ... auth* /app/` and the later `COPY . /app`. The build does not need the file: credentials for the Composer install come from `COMPOSER_AUTH`, assembled inside the `RUN` from the `package_token` build secret or the `PACKAGE_TOKEN` build arg. The file is now excluded from the build context and dropped from the copy, so it can no longer reach the image, and a test asserts that it does not.

## Changes

- Added `auth*` to `.dockerignore` under a new credentials group, placed to follow the header's deny-list ordering (VCS internals, then secrets).
- Removed `auth*` from the `COPY composer.json composer.* patches.lock.* .env* /app/` instruction in `.docker/cli.dockerfile`.
- Updated the `str_replace()` needle in `.vortex/tests/phpunit/Traits/SutTrait.php` that anchors the test-only tooling `COPY` injection to this instruction, so the injection keeps matching.
- Extended `testDockerComposePackageToken()` in `.vortex/tests/phpunit/Functional/DockerComposeWorkflowTest.php` to write an `auth.json` into the build context and assert that the built image does not contain `/app/auth.json`.
- Regenerated installer fixtures via `ahoy update-snapshots`: the baseline `.dockerignore` and `cli.dockerfile` carry the change, and the per-scenario diff fixtures shift their hunk offsets accordingly.

Adding `auth.json` to the template's `.gitignore` is a possible follow-up not included here.

## Test coverage

`testDockerComposePackageToken()` already covered the credential path itself: it requires `drevops/test-private-package` (a private repository), asserts that the image build fails without `PACKAGE_TOKEN`, and asserts that it succeeds with it. Because the system under test is a freshly installed template that carries no `auth.json`, that pair of builds already demonstrated that a private package resolves during the build from the `package_token` build secret alone.

What was not covered is the exclusion this change introduces. The test now writes an `auth.json` into the build context before both builds and asserts that the file is absent from the resulting image. Its token is deliberately invalid, which leaves both existing build outcomes unchanged: the no-token build still fails, and the token build still succeeds because Composer re-applies `COMPOSER_AUTH` over any project-level `auth.json`. The only condition the new assertion can fail on is `auth*` no longer being excluded from the build context.

That isolation was confirmed by removing the `.dockerignore` entry and re-running the test: both builds behaved exactly as before and only the new assertion failed.

## Before / After

```
Before
┌──────────────────────┐  COPY ... auth* /app/  ┌──────────────────────┐
│ Build context        │ ──────────────────────▶│ cli image            │
│ (auth.json included) │  COPY . /app           │ /app/auth.json       │
│                      │ ──────────────────────▶│ (token baked in)     │
└──────────────────────┘                        └──────────────────────┘

After
┌──────────────────────┐  COPY ... .env* /app/  ┌──────────────────────┐
│ Build context        │ ──────────────────────▶│ cli image            │
│ (auth* excluded by   │  COPY . /app           │ no auth.json         │
│  .dockerignore)      │ ──────────────────────▶│                      │
└──────────────────────┘                        └──────────────────────┘

Both: --mount=type=secret,id=package_token ─▶ COMPOSER_AUTH ─▶ composer install
```

```
Test guard

Before                                    After
  build without token  ─▶ fails             plant auth.json (invalid token)
  build with token     ─▶ succeeds            build without token  ─▶ fails      (unchanged)
  image contents       ─▶ unchecked           build with token     ─▶ succeeds   (unchanged)
                                              /app/auth.json       ─▶ absent     (new)
```


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.56% (206/209)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.58% (209/212)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.58% (209/212)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.58% (209/212)
```
<details>

```
```
</details>

PR #3099, opened 2026-09-08
[#3098] Enabled email rerouting in every cloud environment and moved 'reroute_email' to the development modules script.

### LINKED ISSUE #3098: Enable email rerouting in all cloud environments and install 'reroute_email' from the development modules script
## Summary

Two related changes to how `reroute_email` behaves outside production.

**Rerouting is off in `stage`**, so outgoing mail from a stage site reaches its real recipients. Stage runs a copy of production content, and any address that survives sanitisation can receive mail triggered by a cron run, a content save, or a UAT session. Rerouting should follow one rule: every cloud environment reroutes, every local environment delivers.

**The module is installed by the demo script**, `scripts/provision-00-enable-demo-modules.sh`, whose own header instructs projects to replace its operations once the site stops relying on demo content. A project that follows that instruction stops installing `reroute_email` on every rebuilt environment, and mail starts reaching real recipients from `dev` and `stage`. The install belongs in `scripts/provision-10-enable-dev-modules.sh`, which exists precisely so that non-production modules survive the demo script being adjusted or removed.

`local` and `ci` keep their current behaviour and continue to deliver to the intended recipient. Local development depends on seeing the real address in the local mail catcher, and CI tests assert on the address a message was sent to.

### Current behaviour

[`web/sites/default/includes/modules/settings.reroute_email.php`](https://github.com/drevops/vortex/blob/main/web/sites/default/includes/modules/settings.reroute_email.php) enables rerouting for every environment outside a deny-list:

```php
if (!in_array($settings['environment'], [ENVIRONMENT_LOCAL, ENVIRONMENT_CI, ENVIRONMENT_STAGE, ENVIRONMENT_PROD], TRUE)) {
  $config['reroute_email.settings']['enable'] = TRUE;
}
else {
  $config['reroute_email.settings']['enable'] = FALSE;
}
```

| Environment      | Rerouting | Mail reaches real recipients |
|------------------|-----------|------------------------------|
| `local`          | off       | no, the local mail catcher intercepts it |
| `ci`             | off       | no, the mail collector stores it |
| `dev`            | on        | no |
| custom / preview | on        | no |
| `stage`          | off       | **yes** |
| `prod`           | off       | yes, intended |

### Proposed behaviour

Drop `ENVIRONMENT_STAGE` from the list and invert the condition, so the environments that deliver real mail are enumerated up front and rerouting becomes the fall-through default:

```php
if (in_array($settings['environment'], [ENVIRONMENT_LOCAL, ENVIRONMENT_CI, ENVIRONMENT_PROD], TRUE)) {
  // Deliver every message to its intended recipient. Local and CI contain mail
  // by other means, and production sends to real recipients.
  $config['reroute_email.settings']['enable'] = FALSE;
}
else {
  // Send every outgoing message to the address above instead of to its
  // intended recipient, unless that recipient matches the allowed list.
  $config['reroute_email.settings']['enable'] = TRUE;
}

// Allow an environment to opt out of the rerouting set above.
if (!empty(getenv('DRUPAL_REROUTE_EMAIL_DISABLED'))) {
  // Deliver every message to its intended recipient.
  $config['reroute_email.settings']['enable'] = FALSE;
}
```

Removing the negation makes the rule readable in one pass: three named environments deliver, everything else reroutes. Any cloud environment type added later reroutes without further changes.

| Environment      | Rerouting | Reason |
|------------------|-----------|--------|
| `local`          | off       | developers need to see the real recipient in the local mail catcher |
| `ci`             | off       | tests assert on the address a message was sent to |
| `dev`            | on        | cloud environment |
| custom / preview | on        | cloud environment |
| `stage`          | on        | cloud environment |
| `prod`           | off       | real delivery is the point |

### On the existing rationale for excluding stage

The documentation gives "Stage may require real email delivery for UAT" as the reason stage is excluded. That case is already covered per-environment by `DRUPAL_REROUTE_EMAIL_DISABLED`, which any environment can set to a non-empty value to force real delivery. A stage site that genuinely needs to send mail to real people should opt in explicitly, rather than every stage site defaulting to it.

## Part 2 - Install the module from the development modules script

`reroute_email` is installed from the `MODULES` block of `scripts/provision-00-enable-demo-modules.sh`, alongside the demo contrib set. That script tells projects to replace its operations once the demo content is gone, and a project that does so silently loses email rerouting on every environment it rebuilds afterwards.

`scripts/provision-10-enable-dev-modules.sh` states the opposite contract in its header: "Development modules are enabled here rather than in the example script so that they remain available after the example script is adjusted or removed." An email safety net is exactly the kind of module that has to survive that.

### Work

- Remove `reroute_email` from the `drush pm:install` line in the `MODULES` fence of `scripts/provision-00-enable-demo-modules.sh`.
- Add a `MODULE_REROUTE_EMAIL` fenced block to `scripts/provision-10-enable-dev-modules.sh` that installs the module, matching the shape of the existing `MODULE_DEVEL` and `MODULE_TESTMODE` blocks.
- Add `reroute_email` to `Modules::DEV_MODULES` in `.vortex/installer/src/Prompts/Handlers/Modules.php`, so the dev script is retained when `reroute_email` is the only one of that set selected. Deselection removal then runs through the existing `File::removeTokenAsync('MODULE_REROUTE_EMAIL')` path rather than the regex that rewrites the demo script's install line, which is hardcoded to `provision-00-enable-demo-modules.sh`.
- Add `$settings['config_exclude_modules'][] = 'reroute_email';` to `settings.reroute_email.php`, matching `settings.devel.php` and the rest of `DEV_MODULES`. Without this the module still reaches production through exported configuration, and the provision script stops being the only install path, which is the property that makes membership of `DEV_MODULES` meaningful. All rerouting settings come from the settings override file, so nothing is lost by keeping the module out of exported configuration.
- Update the `DEV_MODULES` docblock, which currently describes the set as "development and testing tools". That no longer covers an environment-aware safety module.

### Environment gate mismatch to resolve

Both provision scripts gate on an allow-list, skipping any environment that is not `local`, `ci`, `dev` or `stage`. The settings file gates on a deny-list, rerouting anything that is not `local`, `ci` or `prod`.

After this change the two disagree about custom environment types: a custom cloud environment would have rerouting enabled in settings while the provision script treats it as production and never installs the module. Either the scripts should gate on "not production" to match, or the mismatch should be documented as intended.

## Affected files

- `web/sites/default/includes/modules/settings.reroute_email.php` - the environment condition, and the `config_exclude_modules` entry.
- `scripts/provision-00-enable-demo-modules.sh` - the `MODULES` install line.
- `scripts/provision-10-enable-dev-modules.sh` - the new fenced block.
- `.vortex/installer/src/Prompts/Handlers/Modules.php` - the `DEV_MODULES` constant and its docblock.
- `tests/phpunit/Drupal/SwitchableSettingsTest.php` - the "Stage: disabled by default" case in `dataProviderRerouteEmail()`.
- `tests/phpunit/Drupal/EnvironmentSettingsTest.php` - the Acquia stage and Lagoon test environment cases, which expect `enable = FALSE`, and the `config_exclude_modules` assertions across every environment case.
- `.vortex/tooling/tests/unit/provision.bats` - the asserted `drush pm:install` line for the demo script, and new assertions for the dev script.
- `.vortex/docs/content/development/modules/contributed-modules.mdx` - the sentence naming the excluded environments, the environment behaviour table, and the row explaining why stage is excluded.
- Installer fixtures under `.vortex/installer/tests/Fixtures/handler_process/` hold copies of both provision scripts, the settings file and `EnvironmentSettingsTest.php`, so `ahoy update-snapshots` needs a run.

AlexSkrypnyk:
Closes #3098

## Summary

`web/sites/default/includes/modules/settings.reroute_email.php` now enables `reroute_email` in every environment except `local`, `ci` and `prod`. `ENVIRONMENT_STAGE` left the list of environments that deliver mail, and the condition lost its negation, so rerouting is the fall-through default rather than the exception; the same file adds `reroute_email` to `$settings['config_exclude_modules']`.

A `stage` environment runs a copy of production content, so before this change any address that database sanitization missed received real mail as soon as a cron run, a content save or a UAT session triggered one. Separately, the module's `drush pm:install` call sat in the `MODULES` fence of `scripts/provision-00-enable-demo-modules.sh`, whose header tells projects to replace its operations once the site stops relying on demo content - a project that follows that instruction stops installing `reroute_email` on every environment it rebuilds afterwards, and mail starts reaching real recipients from `dev` as well.

`reroute_email` now installs from a `MODULE_REROUTE_EMAIL` block in `scripts/provision-10-enable-dev-modules.sh`, placed ahead of `MODULE_GENERATED_CONTENT` so the safety net is in place before content generation creates users, and it joins `Modules::DEV_MODULES` so the installer keeps that script when `reroute_email` is the last selected member of the set. `local` and `ci` still deliver to the intended recipient, and `DRUPAL_REROUTE_EMAIL_DISABLED` still forces delivery on any single environment - that is the supported route for a `stage` site that genuinely needs to mail real people during UAT. Both provision scripts keep their `local`/`ci`/`dev`/`stage` gate, and `provision.mdx` now documents the resulting disagreement with the settings file over custom environment types instead of resolving it.

## Before / After

```
BEFORE
├─ settings.reroute_email.php
│    if (!in_array($env, [LOCAL, CI, STAGE, PROD], TRUE))  enable = TRUE
│    else                                                  enable = FALSE
│
│                 local   ci    dev   stage   prod   custom
│    rerouting      off   off   ON     off     off     ON
│                                       ▲
│                                       └── delivers to real inboxes
│
└─ provision-00-enable-demo-modules.sh   (projects are told to replace it)
        drush pm:install ... redirect reroute_email robotstxt ...

AFTER
├─ settings.reroute_email.php
│    $settings['config_exclude_modules'][] = 'reroute_email';
│
│    if (in_array($env, [LOCAL, CI, PROD], TRUE))          enable = FALSE
│    else                                                  enable = TRUE
│
│                 local   ci    dev   stage   prod   custom
│    rerouting      off   off   ON     ON      off     ON
│                                       ▲
│                                       └── reroutes, like every other
│                                           cloud environment
│
├─ provision-00-enable-demo-modules.sh
│      drush pm:install ... redirect robotstxt ...
│
└─ provision-10-enable-dev-modules.sh    (exists to survive that replacement)
        #;< MODULE_REROUTE_EMAIL
        drush pm:install reroute_email
        #;> MODULE_REROUTE_EMAIL
```

## Changes

**Behaviour**

- `web/sites/default/includes/modules/settings.reroute_email.php`: dropped `ENVIRONMENT_STAGE` from the environments that deliver mail and inverted the condition so rerouting is the fall-through default; added `reroute_email` to `$settings['config_exclude_modules']` so the module cannot reach production through exported configuration.
- `scripts/provision-00-enable-demo-modules.sh`: removed `reroute_email` from the contrib `drush pm:install` line.
- `scripts/provision-10-enable-dev-modules.sh`: added a `MODULE_REROUTE_EMAIL` fenced block that installs the module, placed ahead of `MODULE_GENERATED_CONTENT`.

**Installer**

- `.vortex/installer/src/Prompts/Handlers/Modules.php`: added `reroute_email` to `Modules::DEV_MODULES`, so deselecting it removes the fenced block through `File::removeTokenAsync()` and the dev script is deleted only once every member of the set is deselected. The constant's docblock described the set as development and testing tools, which no longer covers an environment-aware safety module, so it now names the shared property instead.
- Fixtures under `.vortex/installer/tests/Fixtures/handler_process/` regenerated with `ahoy update-snapshots`, including the new `modules_no_devel_sdc_devel_generated_content_testmode_reroute_email` scenario.

**Tests**

- `tests/phpunit/Drupal/SwitchableSettingsTest.php`: the stage case now expects `enable = TRUE`, and a new case covers `stage` with `DRUPAL_REROUTE_EMAIL_DISABLED` set.
- `tests/phpunit/Drupal/EnvironmentSettingsTest.php`: `testEnvironmentAcquiaStage` and `testEnvironmentLagoonTest` now expect `enable = TRUE`, and the fenced `reroute_email` entry was added to all 16 `config_exclude_modules` assertions.
- `.vortex/tooling/tests/unit/provision.bats`: the install assertion moved from the demo-modules block to the dev-modules block across all 13 scenarios that install development modules.
- `.vortex/tooling/tests/unit/provision-enable-dev-modules.bats`: added the `reroute_email` drush mock expectation and its output assertions to both installing scenarios, and the not-expected line to the two production-skip scenarios.
- `.vortex/installer/tests/Functional/Prompts/Handlers/ModulesHandlerProcessTest.php`: `modules_no_reroute_email` now asserts the dev script survives without the install line, `modules_no_devel_sdc_devel_generated_content_testmode` asserts it survives on `reroute_email` alone, and a new scenario asserts it is deleted only when all five members are deselected.

**Documentation**

- `.vortex/docs/content/development/modules/contributed-modules.mdx`: added the install-source bullet, flipped the stage row in the environment behaviour table, recorded the config export exclusion, and named `DRUPAL_REROUTE_EMAIL_DISABLED` as the supported stage UAT opt-out.
- `.vortex/docs/content/development/provision.mdx`: added `reroute_email` to the dev script's "present when" list and to the section describing it, and documented that both provision scripts skip any environment type outside `local`, `ci`, `dev` and `stage` while the settings file reroutes everything outside `local`, `ci` and `prod`.
- `.vortex/docs/content/development/modules/README.mdx`: the development provisioning and config-exclusion rows no longer describe the set as development-only.

**Not included**

The recorded terminal demos (`.vortex/docs/static/img/provision.*` and siblings) still show the old demo install line. Re-recording rebuilds the whole set with timing-dependent output, CI does not verify them, and they are re-recorded wholesale during release preparation.


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.59% (210/213)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.59% (210/213)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.59% (210/213)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.59% (210/213)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.59% (210/213)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.59% (210/213)
```
<details>

```
```
</details>

PR #3118, opened 2026-09-08
[#3105] Matched the prod domain placeholder case in the Acquia purge-cache task.

### LINKED ISSUE #3105: Match the prod domain placeholder case in the Acquia purge-cache task
## Summary

`vortex-task-purge-cache-acquia` never strips the environment placeholder from production domains, because the strip patterns are lowercase while the shipped placeholder is uppercase. Bash pattern matching is case-sensitive, so the replacement silently does nothing and a production domain keeps an `prod.` prefix that was meant to be removed.

## Details

`hooks/library/domains.txt` ships a single placeholder line:

```
$TARGET_ENV_REMAP.your-site-domain.example
```

`.vortex/tooling/src/vortex-task-purge-cache-acquia` builds the domain list from that file. For the production environment it tries to drop the placeholder entirely:

```bash
# Strip placeholder for PROD environment.
if [ "${target_env}" = "prod" ]; then
  domain="${domain//\$target_env_remap./}"
  domain="${domain//\$target_env./}"
fi
```

Both patterns are lowercase (`$target_env_remap.`, `$target_env.`), but the text in the domains file is uppercase (`$TARGET_ENV_REMAP.`). `${var//pattern/}` is case-sensitive, so neither replacement matches.

The placeholder therefore survives to the interpolation step:

```bash
domain="$(eval echo "${domain}")"
```

which expands `$TARGET_ENV_REMAP` against the shell variable of the same name. For production that variable holds `prod`, so the resulting domain is `prod.your-site-domain.example` rather than the intended `your-site-domain.example`. The cache purge then targets a domain that may not exist for the site.

**Expected:** for `prod`, the placeholder segment is removed and the bare domain is used.
**Actual:** the placeholder is expanded and the domain keeps a `prod.` prefix.

### Note for whoever picks this up

Renaming the shell variable to lowercase looks like the obvious fix and is not safe. The `eval` on the domain string resolves `$TARGET_ENV_REMAP` by name against the shell variable, so lowercasing the variable while the shipped `domains.txt` keeps the uppercase token would make the expansion produce an empty segment and break every environment, not just production. Either the strip patterns move to uppercase, or the placeholder token and the variable are changed together.

### Test coverage

There is no BATS suite for this script - `.vortex/tooling/tests/unit/` has no `task-purge-cache-acquia.bats`. That is why the mismatch has gone unnoticed. A regression test covering the `prod` branch would be worth adding alongside the fix.

AlexSkrypnyk:
Closes #3105

## Summary

`.vortex/tooling/src/vortex-task-purge-cache-acquia` now strips the `$TARGET_ENV_REMAP.` and `$TARGET_ENV.` placeholder tokens from the domain it compiles for the `prod` environment using case-matched patterns, so the domain purged for production is `your-site-domain.example` instead of `prod.your-site-domain.example`.

The two `${domain//pattern/}` strip patterns in the `prod` branch were lowercase (`\$target_env_remap.`, `\$target_env.`) while `hooks/library/domains.txt` ships the uppercase token `$TARGET_ENV_REMAP.your-site-domain.example`, and Bash pattern substitution is case-sensitive, so when `target_env` is `prod` neither pattern matched and the placeholder reached `domain="$(eval echo "${domain}")"`, which expands `$TARGET_ENV_REMAP` by name against the shell variable set to `prod` a few lines earlier.

The Acquia purge for `prod` now targets the real production domain instead of `prod.your-site-domain.example`, a host that typically does not exist and that the task reported as a skipped missing domain while still exiting successfully; the `dev`, `test`, and `test2` branches, which intentionally keep their environment prefix, are untouched, and `.vortex/tooling/tests/unit/task-purge-cache-acquia.bats` adds the script's first BATS coverage so a future token-case change fails the suite instead of passing against a stale fixture.

## Before / After

```
BEFORE: lowercase strip patterns miss the shipped token
┌─────────────────────────────────────────────────────────────────┐
│ domains.txt   $TARGET_ENV_REMAP.your-site-domain.example        │
│ strip          ${domain//\$target_env_remap./}  (no match)      │
│ eval           expands $TARGET_ENV_REMAP -> prod                │
│ purged         prod.your-site-domain.example (missing, skipped) │
└─────────────────────────────────────────────────────────────────┘

AFTER: uppercase strip patterns match the shipped token
┌─────────────────────────────────────────────────────────────────┐
│ domains.txt   $TARGET_ENV_REMAP.your-site-domain.example        │
│ strip          ${domain//\$TARGET_ENV_REMAP./}  (stripped)      │
│ eval           nothing left to expand                           │
│ purged         your-site-domain.example (real host, purged)     │
└─────────────────────────────────────────────────────────────────┘
```

## Changes

- Uppercased both strip patterns in the `prod` branch of `.vortex/tooling/src/vortex-task-purge-cache-acquia` to `$TARGET_ENV_REMAP.` and `$TARGET_ENV.`, matching the case of the token `hooks/library/domains.txt` ships and the legacy pre-2022 token some consumer copies may still carry.
- Kept `TARGET_ENV_REMAP`/`TARGET_ENV` uppercase rather than renaming the shell variables to lowercase: the later `eval echo "${domain}"` resolves the placeholder by name against those variables, so lowercasing them would expand to an empty segment on every environment, not just `prod`.
- Added `.vortex/tooling/tests/unit/task-purge-cache-acquia.bats` with 8 tests driven by the real shipped `hooks/library/domains.txt`: `prod` strips the shipped placeholder, `prod` strips the legacy `$TARGET_ENV.` token, `dev` / `test` (remapped to `stage`) / `test2` keep their prefixes, an unknown environment compiles no domains, a full `prod` purge completes through the notification-status poll, and the missing-key guard stops before any request.
- Every `curl` call in the new suite is mocked through the `STEPS`/`steps_run` idiom used by `fetch-db-acquia.bats`, so the tests run without network access.

## Screenshots

N/A

### DIFF
--- a/.vortex/tooling/src/vortex-task-purge-cache-acquia
+++ b/.vortex/tooling/src/vortex-task-purge-cache-acquia
@@ -100,8 +100,8 @@ while read -r domain; do
   TARGET_ENV_REMAP="${target_env}"
   # Strip placeholder for PROD environment.
   if [ "${target_env}" = "prod" ]; then
-    domain="${domain//\$target_env_remap./}"
-    domain="${domain//\$target_env./}"
+    domain="${domain//\$TARGET_ENV_REMAP./}"
+    domain="${domain//\$TARGET_ENV./}"
   fi
 
   # Re-map 'test' to 'stage' as seen in UI.
--- a/.vortex/tooling/tests/unit/task-purge-cache-acquia.bats
+++ b/.vortex/tooling/tests/unit/task-purge-cache-acquia.bats
@@ -1,0 +1,215 @@
+#!/usr/bin/env bats
+##
+# Unit tests for the Acquia cache purge task.
+#
+# shellcheck disable=SC2030,SC2031,SC2034
+
+load ../_helper.bash
+
+# Runs the task for one environment and asserts the domain it purges.
+#
+# The Acquia API replies without a notification link, which the task reports as
+# a missing domain and skips, so the run stops before the status polling.
+#
+# Arguments:
+#   1. env_name: Environment to purge the cache for.
+#   2. domain: Domain the placeholder is expected to resolve to.
+#   3. domains_file: File to read the domain list from. Optional, defaults to
+#      the domain list shipped with the template.
+assert_purge_domain() {
+  local env_name="${1}"
+  local domain="${2}"
+  local domains_file="${3:-hooks/library/domains.txt}"
+
+  declare -a STEPS=(
+    "[INFO] Started cache purging in Acquia."
+
+    "Retrieving authentication token."
+    '@curl -s -L https://accounts.acquia.com/api/auth/oauth/token --data-urlencode client_id=test-key --data-urlencode client_secret=test-secret --data-urlencode grant_type=client_credentials # {"access_token":"test-token"}'
+    "[ OK ] Retrieved authentication token."
+
+    "Retrieving testapp application UUID."
+    '@curl -s -L -H Accept: application/json, version=2 -H Authorization: Bearer test-token https://cloud.acquia.com/api/applications?filter=name%3Dtestapp # {"_embedded":{"items":[{"uuid":"app-uuid-123"}]}}'
+    "[ OK ] Retrieved testapp application UUID."
+
+    "Retrieving environment ID."
+    "@curl -s -L -H Accept: application/json, version=2 -H Authorization: Bearer test-token https://cloud.acquia.com/api/applications/app-uuid-123/environments?filter=name%3D${env_name} # {\"_embedded\":{\"items\":[{\"id\":\"env-id-456\"}]}}"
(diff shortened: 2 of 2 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.68% (224/227)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.68% (224/227)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.68% (224/227)
```
<details>

```
```
</details>

PR #3119, opened 2026-09-08
[#3106] Added the missing marker space in negative BATS step assertions.

### LINKED ISSUE #3106: Add the missing space in negative step assertions in 'notify-github.bats'
## Summary

Three negative assertions in `notify-github.bats` are written without a space after the leading dash, so the steps parser strips two characters from the wrong offset and the tests assert the absence of a truncated string. They pass, but not for the reason they appear to.

## Details

The `STEPS` array uses a leading `-` to mean "this text must be absent from the output". The bats-helpers steps parser removes exactly two leading characters (the marker and the space that follows it) before matching.

In `.vortex/tooling/tests/unit/notify-github.bats` three entries omit the space:

- line 115: `"-Marked deployment as finished."`
- line 213: `"-Marked deployment as finished."`
- line 314: `"-Marked deployment as finished."`

After the parser removes two characters, the asserted string is `arked deployment as finished.` - the leading `M` is consumed along with the dash. The assertion still holds, because output that does not contain `Marked deployment as finished.` also does not contain `arked deployment as finished.`, so the tests are green for the wrong reason.

The dominant form elsewhere in the suite includes the space:

```bash
"- Marked deployment as finished."
```

`provision.bats` has a related variant where alignment spaces are embedded in the asserted text (for example `"-      Existing database detected..."`), which is worth reviewing at the same time.

**Expected:** the negative assertion checks the absence of `Marked deployment as finished.`
**Actual:** it checks the absence of `arked deployment as finished.`

Adding the missing space in the three entries is enough; the assertions keep passing and start asserting the intended string.

AlexSkrypnyk:
Closes #3106

## Summary

Negative step assertions in `notify-github.bats` and `provision.bats` now write the `-` marker followed by a space before the asserted substring, the separator `steps_run()`'s `${item:2}` slice assumes is always there.

`steps_run()`'s guard `[[ ${item} == "-"* ]]` matches a bare dash and `${item:2}` then unconditionally drops two characters, so a step written without the space loses its first content character; the three `notify-github.bats` entries read `"-Marked deployment as finished."`, so the suite asserted the absence of `arked deployment as finished.` and could never detect a regression in the real message, while the sibling branch's guard `[[ ${item} == "= "* ]]` checks for the space directly and never degrades this way.

In `provision.bats` the same missing marker space sat on 11 negative `Existing database detected` assertions, but there it consumed a leading space rather than a letter, so `assert_output_not_contains` matched the intended substring either way and that file's change is formatting-only: those 11 rows, together with the 2 positive `Existing` rows that carried no alignment at all, now use the 6-space indent the file's `Fresh database detected` rows already had. `steps_run()` in `bats-helpers` and every other `.bats`/`.bash` file in the repo are unchanged.

## Before / After

```
BEFORE - marker written without a space
──────────────────────────────────────────────────────
STEPS entry     "-Marked deployment as finished."
                       │
                       │  steps_run(): ${item:2}
                       ▼
asserted text   "arked deployment as finished."
                       │
                       ▼
assert_output_not_contains → always true, the real
message is never checked

AFTER - marker written with a space
──────────────────────────────────────────────────────
STEPS entry     "- Marked deployment as finished."
                       │
                       │  steps_run(): ${item:2}
                       ▼
asserted text   "Marked deployment as finished."
                       │
                       ▼
assert_output_not_contains → fails if the real message
reappears in the output
```

## Changes

1. `.vortex/tooling/tests/unit/notify-github.bats` - added the missing space after the `-` marker in the 3 negative `Marked deployment as finished.` assertions covering the pre-deployment, post-deployment, and deployment-status-update failure paths.
2. `.vortex/tooling/tests/unit/provision.bats` - added the same marker space to the 11 negative `Existing database detected...` assertions and aligned them to the file's 6-space indent; realigned the 2 positive `Existing database detected...` assertions to match their `Fresh database detected...` counterparts. No assertion changes meaning; this file's change is formatting-only.

## Screenshots

N/A

### DIFF
--- a/.vortex/tooling/tests/unit/notify-github.bats
+++ b/.vortex/tooling/tests/unit/notify-github.bats
@@ -112,7 +112,7 @@ load ../_helper.bash
     '@curl -X POST -H Authorization: token token12345 -H Accept: application/vnd.github.v3+json -s https://api.github.com/repos/myorg/myrepo/deployments -d {"ref":"nonexistingbranch","environment":"nonexistingbranch","auto_merge":false,"required_contexts":[]} # {"message": "No ref found for: nonexistingbranch","documentation_url": "https://docs.github.com/rest/deployments/deployments#create-a-deployment","status": "422"}'
     "Unable to get a deployment ID for a pre_deployment operation. Payload:"
     "Wait for GitHub checks to finish and try again."
-    "-Marked deployment as finished."
+    "- Marked deployment as finished."
   )
 
   mocks="$(steps_run "setup")"
@@ -210,7 +210,7 @@ load ../_helper.bash
     "@curl -X GET -H Authorization: token token12345 -H Accept: application/vnd.github.v3+json -s https://api.github.com/repos/myorg/myrepo/deployments?ref=nonexistingbranch # []"
     "Unable to get a deployment ID for a post_deployment operation. Payload:"
     "Check that a pre_deployment notification was dispatched."
-    "-Marked deployment as finished."
+    "- Marked deployment as finished."
   )
   mocks="$(steps_run "setup")"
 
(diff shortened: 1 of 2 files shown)


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.68% (224/227)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.68% (224/227)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   98.68% (224/227)
```
<details>

```
```
</details>

Issue #2953, opened 2026-08-11
Rename inconsistent consumer-facing environment variables and script names in 2.x


AlexSkrypnyk:
## Summary

A full convergence pass over the codebase (PR #2952) catalogued a set of consumer-facing naming inconsistencies across `VORTEX_*` / `DRUPAL_*` environment variables and the shipped tooling script filenames. They were deliberately **not** renamed in that PR: every one of these names is a public contract that consumers set in their own `.env` and hosting configuration, or invoke as `vendor/bin/vortex-*`. Renaming them on `main` would break downstream projects silently - no error, just a setting that stops taking effect.

They belong in `2.x`, where a deprecation-alias cycle can land: accept the new name, keep the old one working with a deprecation notice, then drop the old name in a later major.

### 1. Missing `TASK_` infix

`VORTEX_PURGE_CACHE_ACQUIA_SKIP` drops the `TASK_` infix that its nine sibling variables and both other task hooks carry. Every other task-family variable reads `VORTEX_TASK_<OPERATION>_<PLATFORM>_*`.

### 2. `LAGOONCLI` vs `CLI` for the same binary

`vortex-deploy-lagoon` and `vortex-task-custom-lagoon` use different tokens for the same Lagoon CLI binary - one spells it `LAGOONCLI`, the other `CLI`.

### 3. Toggle polarity split three ways

Feature toggles are spelled `_ENABLED` (opt-in), `_DISABLED` (opt-out) and `_SKIP` (opt-out) across the codebase. A consumer cannot infer the polarity of a new toggle from its name.

### 4. Notification destination and credential naming

The destination variable is named three ways for the same concept:

| Channel | Variable suffix |
|---|---|
| JIRA, New Relic | `_ENDPOINT` |
| Webhook | `_URL` |
| Slack | bare `_WEBHOOK` |

Credentials split similarly between `_TOKEN` (Diffy, GitHub, JIRA) and `_USER_KEY` (New Relic).

### 5. Asymmetric admin block/unblock variables

`VORTEX_LOGIN_UNBLOCK_ADMIN` and `VORTEX_LOGOUT_BLOCK_ADMIN` encode the same underlying choice from opposite directions, where a single generic fallback would serve both.

### 6. `VORTEX_FETCH_DB_FORCE` vs `VORTEX_FETCH_DB_FRESH`

These two names are near-identical but have genuinely different semantics:

- `VORTEX_FETCH_DB_FORCE` - skip reuse of the local cached copy.
[...]

PR #3150, opened 2026-09-17
[#3135] Added an opt-in check that fails provisioning when post-provision operations change configuration.

### LINKED ISSUE #3135: Fail provisioning when exported configuration does not match the database
## Summary

Provisioning imports configuration but never verifies that the configuration in code and the active configuration agree once all post-provision operations have finished. A site can finish provisioning successfully while carrying configuration that was written to the database after the import and never exported. The drift stays invisible until someone runs an export weeks later and finds an unrelated diff they have to untangle.

### How this differs from the existing check

`VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_UPDATE` already exists and must not be confused with this. That check exports configuration before and after `drush updatedb`, diffs the two directories, and fails when update hooks changed active configuration - so it catches one specific cause, and it runs *before* `drush config:import`.

The gap is everything that changes active configuration *after* that point. Reading the script in order, all of these run after the import and are currently unverified:

- `drush config:import`, its optional repeat, and the extra `config_split` import - an import that does not converge leaves drift behind.
- `drush deploy:hook`.
- The custom provision scripts loop over `scripts/provision-*.sh`. This is the significant one: the shipped examples enable modules and run migrations, and enabling a module writes configuration.

So the two checks are complementary, not alternatives: the existing one asks "did update hooks change config before we imported", and this one asks "after everything ran, does the site still match what is in code".

### Where the check goes

In `.vortex/tooling/src/vortex-provision`, at the very end of the post-provision operations: **after the maintenance-mode disable block (`if [ "${VORTEX_PROVISION_USE_MAINTENANCE_MODE}" = "1" ]` ... `drush maint:set 0`, currently around lines 430-435) and immediately before the `duration=$(($(date +%s) - start_time))` calculation and the closing `pass` (currently around lines 437-438).**

Three reasons for that exact position:

1. It is after every operation that can write configuration - the imports, `drush deploy:hook`, database sanitization and the custom provision scripts. Placing it any earlier misses the custom scripts, which are the most likely cause of drift on a real project.
2. It is after maintenance mode is disabled, so a detected drift fails the build without leaving the site in maintenance mode. Config drift is a code-quality failure, not a reason to take a provisioned site offline.
3. It is inside the region guarded by the early return for `VORTEX_PROVISION_POST_OPERATIONS_SKIP`, so the check is correctly skipped for the database-caching path in CI that deliberately stops before post-provision operations.

### Implementation notes

- `drush config:status` is the natural command. `.vortex/tests/phpunit/Traits/Subtests/SubtestAhoyTrait.php` already asserts on its output, both that a clean build reports no `Different` items and that a deliberately changed site reports them, so the command's behaviour in the provisioned environment is already proven.
- Decide deliberately which states fail. `Different` clearly should. `config:status` also reports items present only in the database or only in the sync directory, and a project that legitimately has database-only configuration would fail on those - so either restrict the check to `Different` or make the state set part of the configuration.
- Sanitization is safe to run before the check: `scripts/sanitize.sql` and `vortex-provision-sanitize-db` only touch `users_field_data`, so they cannot produce false drift.
- The new variable should be named consistently with the existing `VORTEX_PROVISION_VERIFY_*` sibling, and the name should be settled against the variable-vocabulary work in #2888 and #2953 rather than chosen in isolation.

## Acceptance criteria

- The check runs at the end of the post-provision operations, in the position described above.
- The check is **opt-in and disabled by default** (`0`), matching the default of the existing `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_UPDATE` sibling, so no existing project changes behaviour on upgrade.
- The check additionally requires configuration files to be present, using the same `site_has_config_files` guard the existing verification uses, so a profile install with no configuration in code does not fail.
- The check is skipped when post-provision operations are skipped.
- On failure the output names the differing configuration items and states what to do about them, following the existing failure block's style of printing the diff and then calling `fail`.
- On success the script emits a `pass` line, consistent with every other step.
- The new setting appears in the `note` information block near the top of provisioning alongside `Verify config after update`, so its state is visible in every provision log.
- The variable name is consistent with the existing `VORTEX_PROVISION_VERIFY_*` sibling.
- Unit and functional coverage for the tooling script covers the enabled-and-clean, enabled-and-drifted, and disabled paths.
- BATS coverage is updated in both `provision.bats` and the relevant per-script provision test file.
- The variable is documented, and the regenerated variables table is committed.

AlexSkrypnyk:
Closes #3135

## Summary

`.vortex/tooling/src/vortex-provision` now runs `drush config:status --state=Different --format=list` after the maintenance-mode-disable block and fails the provision when the opt-in `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_PROVISION` flag (default `0`) is `1` and `site_has_config_files` is true.
The existing `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_UPDATE` check only diffs configuration around `drush updatedb` and runs before `drush config:import`, so configuration written afterwards by `drush deploy:hook`, database sanitization, or a `scripts/provision-*.sh` custom script went unverified and a site could finish provisioning with active configuration never exported to code.
Enabling the new flag makes provisioning exit non-zero and print the `Different`-state config keys when active configuration diverges from the sync directory at the end of the run; items reported as `Only in DB` or `Only in sync dir` do not fail the check, and the flag stays `0` by default so existing pipelines are unaffected.

## Before / After

```
BEFORE
──────
  ⑥ Disable maintenance mode
   ▼
  🏁 End (exit 0)
     ↳ config:import / deploy:hook / sanitize / provision-*.sh
       may have left active config different from the sync dir - never checked

AFTER
─────
  ⑥ Disable maintenance mode
   ▼
  ⑪ Verify config unchanged? (VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_PROVISION)
   │
   ├── Different state found ──► 🏁 EXIT 1 (fail)
   │
   └── Unchanged / check disabled ──► 🏁 End (exit 0)
```

## Changes

- `.vortex/tooling/src/vortex-provision`: declares `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_PROVISION` (default `0`), reports it in the provisioning info block, and runs the `drush config:status --state=Different --format=list` check after maintenance mode is disabled and before the duration calculation, inside the region guarded by the `VORTEX_PROVISION_POST_OPERATIONS_SKIP` early exit.
- `.env`: adds the template block for `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_PROVISION=0` alongside the sibling `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_UPDATE` variable.
- `.vortex/tooling/tests/unit/provision.bats`: adds `Provision: DB; no site; configs; verify config after provision unchanged` and `Provision: DB; no site; configs; verify config after provision changed`, and folds negative assertions plus a `Verify config after provision  : No` check into the existing `Provision: DB; no site; configs` scenario for the disabled path.
- `.vortex/tests/phpunit/Traits/Subtests/SubtestAhoyTrait.php`: runs `ahoy provision` with `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_PROVISION=1` in the workflow subtest and asserts the pass message.
- `.vortex/docs/content/development/provision.mdx`: adds the `⑪ Verify config unchanged?` node to the flow diagram and describes the flag as list item 11.
- `.vortex/docs/content/development/variables.mdx` and the installer `.env` fixtures under `.vortex/installer/tests/Fixtures/handler_process/`: regenerated to include the new variable.

## Notes for the reviewer

- The name follows the existing `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_*` family rather than introducing a new shape, because #2888 and #2953 defer the `VORTEX_*` vocabulary redesign to `2.x`.
- Only the `Different` state fails, and the state set is not configurable - one new variable rather than two. A project that legitimately carries database-only or sync-directory-only configuration is unaffected.
- `variables.mdx` reports the default as `UNDEFINED`. That is how Shellvar renders every variable defined in both `.env` and a tooling script, including the existing `VORTEX_PROVISION_VERIFY_CONFIG_UNCHANGED_AFTER_UPDATE`, `VORTEX_PROVISION_OVERRIDE_DB` and `VORTEX_PROVISION_SANITIZE_DB_SKIP`. The table is generated, so hand-editing that cell would be reverted by `ahoy update-docs` and would fail the docs-freshness check.
- The recorded `provision` demo under `.vortex/docs/static/img/` is one line stale, since the provisioning info block gained a `Verify config after provision` row. It is not verified by CI and re-recording needs a full workspace bootstrap, so it is left to the release checklist, which regenerates every video.

## Summary by CodeRabbit

* **New Features**
  * Added optional post-provision configuration verification.
  * When enabled, provisioning checks that active configuration matches the exported configuration after deployment tasks and reports any differences.
  * Provisioning fails when configuration changes are detected and confirms successful verification when no changes are found.
  * The setting is disabled by default and can be enabled through the new configuration variable.

* **Documentation**
  * Added guidance for configuring post-provision verification and clarified related provisioning options.


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

PR #3151, opened 2026-09-17
[#3139] Added an npm dependency audit to the security audit job.

### LINKED ISSUE #3139: Audit npm dependencies for security advisories
## Summary

The security audit job audits PHP dependencies and scans for committed secrets, but nothing audits the JavaScript dependency tree. The template ships a lock file with a substantial development dependency tree - linters, formatters, the test runner and their transitive dependencies - and advisories against any of those packages are invisible to the pipeline today.

## Details

`.github/workflows/audit.yml` runs `composer audit --locked` and a Gitleaks scan. `.circleci/config.yml` runs the same two checks. Neither runs an equivalent audit for `package.json` and `package-lock.json`, and neither audits the theme's own JavaScript dependencies.

This is half of the dependency attack surface going unwatched in a job whose entire purpose is to watch the other half. The job already has the right shape for it: a failure toggle variable, a container with the tooling available, and a place in both pipelines.

Both dependency sets need covering - the root `package.json` and the theme's `package.json` - since a project's front-end dependencies usually live in the theme.

### Name and semantics

```
VORTEX_CI_NPM_AUDIT_IGNORE_FAILURE
```

Default: undefined, which means the audit fails the build. Set to `1` to let the build pass while the advisory is being addressed.

The existing CI variables use a precise two-word vocabulary, and this needs the right half of it:

- `*_IGNORE_FAILURE` means "run the check, report its findings, but do not fail the build" - used by `VORTEX_CI_COMPOSER_AUDIT_IGNORE_FAILURE`, `VORTEX_CI_GITLEAKS_IGNORE_FAILURE`, `VORTEX_CI_PHPCS_IGNORE_FAILURE`, `VORTEX_CI_PHPSTAN_IGNORE_FAILURE` and fifteen others.
- `*_SKIP` means "do not run it at all" - used by `VORTEX_CI_TEST_RESULTS_SKIP` and `VORTEX_CI_CODE_COVERAGE_PR_COMMENT_SKIP`.

`IGNORE_FAILURE` is the correct choice here, for the same reason the PHP audit uses it: when a team is knowingly carrying an unfixable transitive advisory, the finding should still appear in the job log so it stays visible and can be re-checked on the next run. A `SKIP` variable would hide the finding entirely, which is the wrong bypass for a security check. One variable, matching its sibling exactly - no second variable for skipping.

### Naming tension worth settling

There is an inconsistency in the existing names that this change has to pick a side on. The audit sibling is named after the tool (`VORTEX_CI_COMPOSER_AUDIT_IGNORE_FAILURE`), while the JavaScript lint variable is named after the ecosystem (`VORTEX_CI_NODEJS_LINT_IGNORE_FAILURE`).

`NPM_AUDIT` is proposed because it names the tool that runs, which pairs it visually with `COMPOSER_AUDIT` - two audits, each named for its package manager. `NODEJS_LINT` is named for the ecosystem because that step runs several tools behind npm scripts, whereas this step runs exactly one command. If reviewers prefer `VORTEX_CI_NODEJS_AUDIT_IGNORE_FAILURE` for consistency with the lint variable instead, that is a reasonable call - it should be settled against the variable-vocabulary work in #2888 and #2953 rather than in isolation.

### Severity threshold is not a new variable

The audit needs a severity floor, or low-severity advisories in transitive development dependencies will block every build. This should not become a second `VORTEX_*` variable: npm reads `audit-level` from `.npmrc`, so the threshold belongs there, in the tool's own configuration, where a project can change it without a template variable existing for it.

### Where it is declared

Like its siblings, this is a CI-provider variable rather than an `.env` variable - `vars.VORTEX_CI_NPM_AUDIT_IGNORE_FAILURE` in GitHub Actions and a project environment variable in CircleCI. Its row in the variables table is sourced from `CI config` with a default of `UNDEFINED`, matching the PHP audit row.

### Documentation to update

The PHP audit variable appears in four places, and the new one needs the same treatment:

- `.vortex/docs/content/development/variables.mdx` - generated, so regenerate with `ahoy update-docs` and commit the result rather than hand-editing the table.
- `.vortex/docs/content/development/security/dependency-audit.mdx` - the how-to that explains setting it to `1`.
- `.vortex/docs/content/continuous-integration/README.mdx` - the table of CI checks and their bypass variables.
- `.vortex/docs/content/development/faqs.mdx` - the entry on bypassing CI for a known advisory.

## Acceptance criteria

- The security audit job audits JavaScript dependencies for known advisories in both CI providers.
- Both the root and the theme dependency sets are audited.
- The audit fails the job on findings at or above a severity threshold configured through `.npmrc`, so low-severity noise in transitive development dependencies does not block every build.
- Failure can be downgraded through `VORTEX_CI_NPM_AUDIT_IGNORE_FAILURE`, defaulting to undefined so the audit fails the build out of the box.
- The finding is still printed when the variable is set, so a tolerated advisory stays visible in the job log.
- No second variable is introduced for skipping the audit or for setting the severity threshold.
- A specific advisory can be excluded with a documented mechanism, matching the escape hatch the PHP audit already has.
- The step is fenced so projects that do not ship JavaScript dependencies do not receive it.
- The variable is documented in all four places listed above, with the variables table regenerated rather than hand-edited.
- Installer fixtures and snapshots are updated for the new CI step.

AlexSkrypnyk:
Closes #3139

## Summary

The `audit` job in `.github/workflows/audit.yml` and the `audit` workflow in `.circleci/config.yml` now run `npm audit --package-lock-only` against the root `package-lock.json` and again with `--prefix="${WEBROOT}/themes/custom/${DRUPAL_THEME}"` against the theme's, both placed between the existing `composer audit --locked` step and Gitleaks and both gated by the new `VORTEX_CI_NPM_AUDIT_IGNORE_FAILURE` variable.

Before this change the job watched only the PHP dependency tree, so an advisory published against a package pinned in the root or theme `package-lock.json` never surfaced in CI regardless of severity. npm's own audit fails on any advisory by default, including low-severity findings in transitive development dependencies, so the new root and theme `.npmrc` files set `audit-level=high` to keep that noise from failing every build.

After merge, a project generated without the `frontend_all` tool group - now carrying the `TOOL_NPM` token - receives neither `.npmrc` nor the root `npm audit` step, while the theme's step stays fenced separately under the existing `DRUPAL_THEME` token, so a project that drops every front-end lint tool but keeps a custom theme still audits it. This does not touch npm's lack of a per-advisory ignore list, so a specific advisory still leaves the tree only via an `overrides` entry in `package.json`.

## Before / After

```
BEFORE - one dependency tree watched, and one check short-circuits the rest:

┌──────────────────────────────┐
│ composer audit --locked      │  on_success  PHP packages (composer.lock)
└──────────────────────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Gitleaks                     │  always      committed secrets
└──────────────────────────────┘

package-lock.json (root) and web/themes/custom/<theme>/package-lock.json
carry no security coverage at all.

AFTER - both dependency trees watched, every check reports in one run:

┌──────────────────────────────┐
│ composer audit --locked      │  always      PHP packages (composer.lock)
└──────────────────────────────┘
               │
               ▼
┌──────────────────────────────┐
│ npm audit (root)             │  always      #;< TOOL_NPM
└──────────────────────────────┘
               │
               ▼
┌──────────────────────────────┐
│ npm audit (theme)            │  always      #;< DRUPAL_THEME
└──────────────────────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Gitleaks                     │  always      committed secrets
└──────────────────────────────┘

The right-hand column is the CircleCI step condition. `always` is spelled
`if: ${{ !cancelled() }}` on the GitHub Actions side, which every step of
that workflow already used.
```

## Changes

- Added an `npm audit --package-lock-only` step for the root lock file and a second one with `--prefix="${WEBROOT}/themes/custom/${DRUPAL_THEME}"` for the theme's, in both `.github/workflows/audit.yml` and `.circleci/config.yml`; the root step is fenced `TOOL_NPM`, the theme step is fenced `DRUPAL_THEME` as a sibling rather than nested, and both honor `VORTEX_CI_NPM_AUDIT_IGNORE_FAILURE`, matching how `VORTEX_CI_COMPOSER_AUDIT_IGNORE_FAILURE` gates the existing Composer step.
- Added `when: always` to the CircleCI `Audit Composer packages` step, which was the one check in that job still on CircleCI's `on_success` default - a failing `composer audit --locked` would have skipped both npm audits while Gitleaks, sitting after them, still ran. All four checks in the job now run independently, which is what `continuous-integration/README.mdx` documents for this workflow and what the GitHub Actions job already did through `if: ${{ !cancelled() }}`.
- Added `.npmrc` at the repo root and in `web/themes/custom/your_site_theme/`, each setting `audit-level=high` so the audit only fails on high or critical advisories. Both files are needed because `--prefix` makes npm read the project configuration from the prefixed directory, and the threshold lives in npm's own configuration rather than in a second `VORTEX_*` variable.
- Changed `.vortex/installer/src/Prompts/Handlers/Tools.php` so the `frontend_all` tool group carries `'token' => 'TOOL_NPM'` and lists `.npmrc` among its `files`; the token is named after the group rather than its member tools (`TOOL_ESLINT_STYLELINT_JEST`) because marker removal matches by substring, and that name would also match the `TOOL_ESLINT` removal, silently deleting the root audit step whenever ESLint alone was deselected.
- Updated the `tools_groups_no_fe_lint*` cases in `.vortex/installer/tests/Functional/Prompts/Handlers/ToolsHandlerProcessTest.php` to assert the `.npmrc` removal and confirm the theme's audit step survives when the root one is removed.
- Regenerated the installer fixtures under `.vortex/installer/tests/Fixtures/handler_process/` to include the new `.npmrc` files, the audit steps, and the step conditions.
- Added a `## JavaScript dependencies` section to `.vortex/docs/content/development/security/dependency-audit.mdx`, moving the existing Composer content under a sibling `## PHP dependencies` heading.
- Updated the security-audit check list and the ignore-failures table in `.vortex/docs/content/continuous-integration/README.mdx`, added an npm-advisory FAQ entry to `.vortex/docs/content/development/faqs.mdx`, declared `VORTEX_CI_NPM_AUDIT_IGNORE_FAILURE=0` in `.vortex/docs/.utils/variables/extra/ci.variables.sh`, and regenerated the corresponding row in `.vortex/docs/content/development/variables.mdx`.


AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

AlexSkrypnyk:
```
  Classes: 100.00% (1/1)
  Methods: 100.00% (2/2)
  Lines:   100.00% (230/230)
```
<details>

```
```
</details>

